PR #86, opened 2015-01-12
Added BrowserSync

ivanmayes:
Added browsersync to the Gruntfile, should automatically start up now.  You can open the same url with different devices/browsers and it will sync your scrolling and auto refresh devices so that you can sync looking at your styles while working on them.

Currently Ios doesnt sync with scrolling, but it will auto refresh with the rest of the devices.

Because of the way the navigation works, it doesn't currently support syncing the moving between pages, there might be a way to solve this though.

I changed the 'connect' hostname to be your public wifi address so that your devices could access patternlab, but left the hostname line commented in if you want to opt out of having your styleguide available over Wifi.

Sorry about the messy Gruntfile commit, my javascript formatter on save reformatted a couple of things and made it messy.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,26 @@
 module.exports = function(grunt) {
 
+    var os = require('os');
+    var ifaces = os.networkInterfaces();
+    var lookupIpAddress = null;
+    for (var dev in ifaces) {
+        if (dev !== "en1" && dev !== "en0") {
+            continue;
+        }
+        ifaces[dev].forEach(function(details) {
+            if (details.family === 'IPv4') {
+                lookupIpAddress = details.address;
+            }
+        });
+    }
+
+    //If an IP Address is passed
+    //we're going to use the ip/host from the param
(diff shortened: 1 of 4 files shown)


bmuenzenmeyer:
Hi @ivanmayes I started playing with this in dev tonight. There are some things I love, and some I'd like to hammer out more.

- [ ] inject new/updated patterns
- [ ] inject updated data.json
- [ ] look into why scrolling is not working
- [ ] look into nav syncing

bmuenzenmeyer:
http://www.browsersync.io/docs/options/

bmuenzenmeyer:
@ivanmayes I've played around with this - and will likely remove it if I cannot get navigation working. It's a pretty heavy dependency to include only to have it missing a large piece of functionality. Either way I will document usage of this and keep it on the roadmap for future consideration.

ivanmayes:
Understood.  Super busy at work, but will keep my branch with it and mess around at some point to try to get the navigation working.

PR #114, opened 2015-04-22
Fix location of paragraph closing tag to wrap citation

laurendorman:


laurendorman:
Thank you, @bmuenzenmeyer. I've done so here.

Sorry about that!

PR #117, opened 2015-04-22
Add 'View All' for each subsection in menus and generate required files

walmokrani:


bmuenzenmeyer:
@walmokrani @ivanmayes just checked this out - it's great! thanks so much for the help :trophy:

PR #126, opened 2015-06-23
adding multi-compiler functionality

scottnath:
Hi,

This is to add the ability to use alternative compiling engines with pattern lab. Specifically, I want to add twig, but made I tried to make my addition versatile enough so that other templating engines could be used. This should open up PL to use multiple template engines in a single PL instance if a developer needs that.

Please let me know what changes I could do. If this is acceptable, I can expand it to include more than just the patterns, but also the surrounding templates.

thanks,
Scott

p.s. my company will be using this in conjunction with our pattern library system:

http://github.com/pattern-library

### DIFF
--- a/builder/patternlab.js
+++ b/builder/patternlab.js
@@ -21,6 +21,7 @@ var patternlab_engine = function(){
   lh = require('./lineage_hunter'),
   pe = require('./pattern_exporter'),
   pa = require('./pattern_assembler'),
+  twig = require('twig'),
   he = require('html-entities').AllHtmlEntities,
   patternlab = {};
 
@@ -88,6 +89,12 @@ var patternlab_engine = function(){
 
       //make a new Pattern Object
       currentPattern = new of.oPattern(subdir, filename);
+
+      //add the relative path to the pattern
+      currentPattern.patternSourcePath = path.join('./source/_patterns',subdir,filename);
+
+      //add the pattern template engine
+      currentPattern.templateEngine = path.extname(filename).replace('.','');
 
       //see if this file has a state
       assembler.setPatternState(currentPattern, patternlab);
@@ -169,9 +176,17 @@ var entity_encoder = new he();
         //extend pattern data links into link for pattern link shortcuts to work. we do this locally and globally
         pattern.data.link = extend({}, patternlab.data.link);
 
-        pattern.patternPartial = renderPattern(pattern.template, pattern.data, patternlab.partials);
+        if(path.extname(pattern.templateEngine) !== 'mustache'){
+          pattern.patternPartial = renderPatternAlt(pattern, patternlab.config);
+        }else{
+          pattern.patternPartial = renderPattern(pattern.template, pattern.data, patternlab.partials);
+        }
       }else{ // Pass global patternlab data
-        pattern.patternPartial = renderPattern(pattern.template, patternlab.data, patternlab.partials);
+        if(path.extname(pattern.templateEngine) !== 'mustache'){
+          pattern.patternPartial = renderPatternAlt(pattern, patternlab.config);
+        }else{
+          pattern.patternPartial = renderPattern(pattern.template, patternlab.data, patternlab.partials);
+        }
       }
 
       //add footer info before writing
@@ -417,6 +432,38 @@ var entity_encoder = new he();
       return mustache.render(name, data);
     }
   }
+
+  /**
+   * Allows rendering of templates from alternative templating engines
+   *
+   * @param {Object}  pattern  details and data for an individual pattern
+   * @param {Object}  config  pattern lab config from config.json
+   *
+   * @return {String}  compiled html
+   */
+  function renderPatternAlt(pattern, config) {
+
+    switch (pattern.templateEngine) {
+      case 'twig':
+        var twigOptions = config.twig;
+        twigOptions.path = pattern.patternSourcePath;
+        
+        var tpl = twig.twig(twigOptions); //read the file with Twig
+        return tpl.render(pattern.data);
+        break;
+      case 'handlebars':
+      case 'hbs':
+        // add hbs compiler
+        break;
+      case 'swig':
+        // add swig compiler
+        break;
+      default:
+        //console.log('Error: no compiler for ' + pattern.templateEngine + 'available.');
+        return fs.readFileSync(pattern.patternSourcePath, 'utf8')
+    }
+  }
+
 
   function addToPatternPaths(bucketName, pattern){
     //this is messy, could use a refactor.
--- a/builder/patternlab.js
+++ b/builder/patternlab.js
@@ -176,13 +176,14 @@ var entity_encoder = new he();
         //extend pattern data links into link for pattern link shortcuts to work. we do this locally and globally
         pattern.data.link = extend({}, patternlab.data.link);
 
-        if(path.extname(pattern.templateEngine) !== 'mustache'){
+        if(pattern.templateEngine && (path.extname(pattern.templateEngine) !== 'mustache')){
+
           pattern.patternPartial = renderPatternAlt(pattern, patternlab.config);
         }else{
           pattern.patternPartial = renderPattern(pattern.template, pattern.data, patternlab.partials);
         }
       }else{ // Pass global patternlab data
-        if(path.extname(pattern.templateEngine) !== 'mustache'){
+        if(pattern.templateEngine && (path.extname(pattern.templateEngine) !== 'mustache')){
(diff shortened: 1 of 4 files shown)


bmuenzenmeyer:
Hi Scott

Will take a look at this soon - love the idea!

Brian

scottnath:
Hi @bmuenzenmeyer 

Made a slight change on the code-viewer template from "mustache" to "template" to keep the tab title generic.

scottnath:
@bmuenzenmeyer do you need this one pointing to `dev` as well?

bmuenzenmeyer:
I've been considering this PR for a while Scott. I do see these, and do not
intend it to become a silent veto as that's not fair to the community that
values and gives back to the project.

While I completely understand the specific goal you are attempting to
achieve with this - I have some reservations. I'll try to be open and
transparent about them with the hopes that you understand. I don't know
much about Twig to be honest, but adding another explicit templating
framework to the core package dependencies adds a lot of weight that the
majority of people will never use. If someone wants to swap out the
compiler, a user can certainly do that - but shipping with two feels like a
mistake.

Also as PL node is essentially a port, it takes its direction from
@dmolsen's Pattern Lab PHP. The shipped default, (at least for v1), is
mustache. I'd welcome Dave to discuss this PR for appropriateness too.

Is there some way we can morph this into a wiki or readme document that
describes how someone would swap the compiler or support more than one? I
am also completely okay with leaving this PR here as a reference.

scottnath:
Morning Brian, thanks for getting back to me.

I totally understand your trepidation in adding bloat to your project, however in this case it is not as simple as swapping out the compiler. The .mustache templates are hard-coded into the PL system - often directly referencing `.mustache` [example](https://github.com/pattern-lab/patternlab-node/blob/master/builder/patternlab.js#L141):
`var patternVariant = new of.oPattern(subdir, currentPattern.fileName + '-' + variantName + '.mustache', variantFileData);`

My attempt was meant to be the start of a new way to set up your files so that this un-wavering reliance on the mustache templating engine could give way to the user choosing their own template compiler according to the needs of whatever system for which they're building patterns. 

For instance, the new function `renderPatternAlt` that would have been added to patternlab.js, [starting at line 445](https://github.com/pattern-lab/patternlab-node/pull/126/files#diff-0dfc6fa9265978698fef365a5e51a730R445), opens the door for not just twig, but really any templating engine a user might desire.

The goal with that function is that it would eventually take over renderPattern, and that the template-compiling functionality would be functions in a separate set of files. Probably something like:
```
./builder/compilers/mustache-compiler.js
./builder/compilers/twig-compiler.js
./builder/compilers/hbr-compiler.js
etc
```
separating the concerns of folder and app structure from template compiling. Also allowing the option of removing a given library from `package.json`, which removes the bloat, but would not then cause errors due to a missing piece.

I think a mistake I made was making this pattern-compiling automatic instead of config-driven, but we needed PL to be able to compile twig and the small code needed for that seemed like a good springboard to the try out the PR process with patternlab-node.

With Drupal 8 going full-twig, Angular's popularity pushing handlebars, etc, it seems you wouldn't want your excellent tool to be stuck with only one option for the kind of templates it can handle. 

Is there an existing roadmap for pattern lab? Are other template engines planned? 

thanks again,
Scott

dmolsen:
Chiming in quickly... PL2 for PHP is supporting Twig primarily because of Drupal 8. This is in addition to Mustache. Basically I've added the notion of "PatternEngines" to the PHP edition so people can swap in their own pattern renderer. An [outdated spec](https://github.com/pattern-lab/the-spec/blob/01d4faca4067aa81ebd78575403b9466144795f5/SPEC.md) and the "[features for v0.8.0](https://github.com/pattern-lab/patternlab-php/issues/51)" issue (_which has turned into PL2_) act as a loose roadmap for the PHP edition.

So I'm definitely in agreement with opening up renderers. I haven't looked at this PR in detail but hopefully the process can be "install bridge to renderer, update config" perspective.

scottnath:
Thanks @dmolsen . Is there a timeframe for PL2?

bmuenzenmeyer:
I can speak for the node port only...

PL2 for Node may never arrive - I haven't even ported all of PL PHP v1 yet. Dave and I have spoken briefly about the future of this repo vs the other efforts, and we'll have to see. Unfortunately I have limited time and therefore the PL port has always had a limited, trailing scope of the larger flagship product. My goal is to close gaps on the PL PHP v1 first, as the issue log tries to emulate. For that reason you may want to check out Dave's projects if Node is not mature enough yet.

We've seen instances of bugs in PL Node being reported on Dave's repo.... which is unfortunate but a product of my lag in delivering and perhaps (thankfully) both projects' 'approachability'. 

I feel a blog post or clarifying roadmap needed from my end. Hopefully this wasn't too rambling. I will try to think through this more tonight.

bmuenzenmeyer:
Morning @scottnath I've read through this again and like the approach! My only reservations would be explicitly shipping it with twig, but I think that's pretty minor. Right now it's on my [roadmap](https://github.com/pattern-lab/patternlab-node/issues/134) as post v1.0.0 functionality but I think that could change for the better.

If you upgrade to v0.10.0 you'll see quite a bit of change (unfortunately the timing of my changes and this PR somewhat collided) - it would help eventual merging if you could adapt this to the new configuration.

Clarifying question @dmolsen - is support for multiple concurrent compilers in the spec or within the scope of PL PHP v1?

geoffp:
Hey, @bmuenzenmeyer and @dmolsen  -- the company I work for also has a clear and present need for something like PatternEngines in patternlab-node. Because we want to operationalize our templates, which are currently a mixture of Handlebars and Underscore (for better or worse), we are going to implement pluggable template renderers, and it needs to be in the Node version because AFAIK, the templating engines we need to support are JS-only.

We really, really want our work to be of benefit to the wider community. We've already done a PoC with template engine plugins on a much older version of PL-node, but after reading this thread, I'm thinking of examining the PatternEngines/PHP architecture and proceeding from v0.10.0 based on that. Does that sound sane? We would of course also want to accommodate the needs of @scottnath.

bmuenzenmeyer:
Hi @geoffp thanks for the continued interest and use of patternlab. Seeing folks weigh in on how and why they wish to use patternlab continues to motivate me. 

As I've said above - my goal is to keep the initial package dependencies small - but support the behaviors mentioned in this PR. I agree that looking at the PHP architecture is somewhere I want to get toward - but for now as you can see by the [roadmap](https://github.com/pattern-lab/patternlab-node/issues/134) I am still playing catch up functionally with v1 of the flagship product. This isn't set in stone, so If you can help me with this I'd appreciate it. @scottnath's branch looks close to me - with the addendum of not shipping with the additional dependency on twig/underscore/handlebars. As far as I am concerned, Mustache will be the default until work continues post v1.0.0

I hope this helps, I welcome welcome *welcome* contributions and couldn't do it all without help. I plan to look at this more as time allows, but please do take a stab at how to build this off of v0.10 and we can take it from there.

scottnath:
Hi all,

I haven't been able to dig back into this and adjust my PR because at work we needed to move forward with our project, so we switched to the php-twig version of Pattern Lab. 

We're building a [pattern-library](/pattern-library/pattern-library) structure, adding in Pattern Lab as a dependency within that structure. All the work we're doing is in js/node, so using the PHP version makes us mix technologies, which is less than ideal. But, we're essentially using Pattern Lab as a prototyping tool only - running all browsersync/sass-compiling/etc from gulp. The only php function we use is `patternlab/core/console --generate`, which compiles from source->public.

In the long run, I'm hoping we can use the node version.  We don't store our patterns inside of Pattern Lab - instead, we do our work outside of PL, then on file changes (watched by Gulp) we transfer the changed files into ./patternlab/source and generate public again, then Gulp triggers BrowserSync. This goes for all file types: patterns, css/sass, js, images, fonts, etc. Therefore, we should be able to switch out the Pattern Lab version fairly easily.

But the ability to use non-mustache languages is paramount. We're setting up for twig now, but we're going to be doing Angular work too, so we'll want to figure out support for ng templating, or at least plain Handlebars. The idea of implementing a PatternEngines system akin to what the php version is doing could definitely help this along and would be a longer-lasting solution than the piecemeal addition of twig I submitted in this PR.

geoffp:
Hey, thanks for connecting, gentlemen!

I'm going to take a stab at it. Hopefully the stars align and I can get something done for review before too long.

bmuenzenmeyer:
I am going to close this PR as work has moved into other PRs. 
Either way, thanks @scottnath for the initial work thinking through this problem!!!

PR #129, opened 2015-07-28
iframe url with 'patterns' bugfix

scottnath:
Hi @bmuenzenmeyer 

Rather than change the regex, I tried to create a larger solution that may help with other odd-url situations.

The default is to use Pattern Lab's original behavior when sending the url to the iframe.

Please let me know what changes you'd like.

thanks,
Scott


bmuenzenmeyer:
@scottnath can you point this at the `dev` branch please?

PR #130, opened 2015-07-28
Patterns url

scottnath:
Hi @bmuenzenmeyer

Rather than change the regex, I tried to create a larger solution that may help with other odd-url situations.

The default is to use Pattern Lab's original behavior when sending the url to the iframe.

Please let me know what changes you'd like.

thanks,
Scott

ps: moved to PR on `dev` branch


bmuenzenmeyer:
Hi @scottnath! I checked this out and it looks good - and I think it will even help me with an instance of patternlab I have live that handles this scenario pretty ungracefully.  

Will be doing some more testing as I can, but I wanted to get this into v0.10.0

THANKS!

scottnath:
Excellent. thanks

PR #141, opened 2015-09-02
watch scss task corrected

afonsopacifer:
Is there any reason for tasks relating to sass be commented ?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -82,7 +82,7 @@ module.exports = function(grunt) {
 			//	options: {
 			//		livereload: true
 			//	},
-			// 	files: ['source/css/**/*.scss', 'public/styleguide/css/*.scss'],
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
Hi @afonsopacifer it's because sass is not intended to be a dependency per some original guidelines I was provided.  That's why the css is shipped with the distributable too.

Considering https://github.com/pattern-lab/patternlab-node/blob/master/package.json, however, it's downloaded anyway. I should consider a better approach to articulate this.

Will try to merge this when I can.

afonsopacifer:
Got it,
It was only a doubt :D
This PR is to get a little mistake in the task that compiles the sass . Sorry for asking in a wrong place :)

bmuenzenmeyer:
@afonsopacifer per [the contributing document](https://github.com/pattern-lab/patternlab-node/blob/master/CONTRIBUTING.md) can you point this to the dev branch? you dont even have to close this one, just change the target

Thanks!

Issue #146, opened 2015-09-10
Please allow more flexible Mustache partial syntax


e2tha-e:
I have a Pattern Lab implementation I'm porting over from PHP. The PHP version allows the following syntax for partials:

* {{> templates-article }}
* {{> 03-templates/02-article }}
* {{> 03-templates/02-article.mustache }}

The Node version only allows the first. The greater flexibility of the PHP version is mandatory for my implementation. I'm submitting a pull request to allow for this. Please review, comment, and merge. Thanks!

bmuenzenmeyer:
This would be a welcome addition @e2tha-e I was not aware the PHP version supported those. In my opinion they are not necessary because the "pattern key" in all instances is `templates-article` and the numerals are just a poorman's ordering device - but by all means PR away! 

:+1: :

e2tha-e:
@bmuenzenmeyer I found the only advantage to the terse partials was terseness. As the project got bigger, knowing the exact paths of partials was a HUGE timesaver. And on top of that, I created a Mustache file browser which allowed you to recurse through many levels of partials, and this also required exact paths.

bmuenzenmeyer:
Merged the PR associated to this as part of release [v0.12.0](https://github.com/pattern-lab/patternlab-node/releases/tag/v0.12.0)

PR #147, opened 2015-09-10
more-flexible-mustache-partial-syntax

e2tha-e:
Updating regexes and key options to allow for explicit paths to be used in Mustache partial tags. Referenced in issue #146


e2tha-e:
@bmuenzenmeyer I think this will do it for the code and unit test. Please review, comment, and merge.

Issue #150, opened 2015-09-20
Want to get pattern~variant.json files to act like they do in patternlab-php


e2tha-e:
I'm submitting a pull request to refine the mergeData function in `pattern_assembler.js`. The short description is so that `pattern~variant.json` files behave as they do in patternlab-php.

As it stands, in patternlab-node, the data in the 2nd param overwrites the data in the 1st param, which in most cases, is what you want. However, this is problematic in `pseudopattern_hunter.js`, because the 1st param is `variantFileData`. What this means is that if a variant data property has the same key as the primary pattern's, the variant will show what's assigned to the primary pattern.

As an example, say `00-homepage~emergency.json` had a property "title" with value "Emergency". 00-homepage~emergency's title will be "Home Page" and not "Emergency". Simply reversing the parameter order won't fix things because then, the non-emergency 00-homepage will show the emergency block.

In order to fix this, I had to reverse the order of parameters in `pseudopattern_hunter.js`. I then had to update the `mergeData` function in `pattern_assembler.js` so it checks that the parameter in obj2 is undefined before copying the value over from ob1.

bmuenzenmeyer:
Merged to `dev`

PR #151, opened 2015-09-20
update mergeData() so that 2nd param has priority over 1st

e2tha-e:
Ref issue #150 
I think the code is fairly well commented and would therefore speak for how it's trying to accomplish what is wanted in the issue. Please review, comment, and merge.


e2tha-e:
Closing because I need to pull against dev and not master

PR #152, opened 2015-09-20
update mergeData() so that 2nd param has priority over 1st

e2tha-e:
Ref issue #150 
I think the code is fairly well commented and would therefore speak for how it's trying to accomplish what is wanted in the issue. Please review, comment, and merge.

### DIFF
--- a/builder/pattern_assembler.js
+++ b/builder/pattern_assembler.js
@@ -170,8 +170,9 @@
     /**
      * Recursively merge properties of two objects.
      *
-     * @param {object} obj1 If obj1 has properties obj2 doesn't, add to obj2.
-     * @param {object} obj2 This object's properties have priority over obj1.
+     * @param {Object} obj1 If obj1 has properties obj2 doesn't, add to obj2.
+     * @param {Object} obj2 This object's properties have priority over obj1.
+     * @returns {Object} obj2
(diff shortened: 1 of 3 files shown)


bmuenzenmeyer:
Hey @e2tha-e thanks for these pull requests - the activity only renews my own interest in the project. I took care of the simpler ones, and will review this one later in the week with a bit more time.

e2tha-e:
@bmuenzenmeyer Much thanks to you as well! Same goes for @dmolsen and @bradfrost. This is really useful software!

bmuenzenmeyer:
Testing this now. It looks good - but it does have a caveat. If the variant datafile overwrites an original property, and that property is an object with children, one must complete overwrite that object. My test case was to add the following to `00-homepage~emergency.json`:

```json
  "hero" : [
    {
      "headline" : {
        "medium" : "EMERGENCY ATOP THE MOUNTAIN"
      }
    }
  ]
```

This of course works, but since I do not overridethe hero image found in `00-homepage.json`, it reverts all the way back to the`data.json` placeholder. 

I don't really see an obvious way around this at midnight- and while not DRY, I think this is a step in the right direction over the current behavior. Merging into `dev` shortly.

Issue #153, opened 2015-09-20
Want ordering of pattern variants to be like in patternlab-php


e2tha-e:
Currently, in patternlab-node, in the toolbar menu, any pattern variants are listed before the primary pattern. (See how Homepage Emergency comes before Homepage.) I would like the primary pattern to be listed first, and the variants after as it is in patternlab-php. Am submitting pull request for this.

bmuenzenmeyer:
Merged

PR #154, opened 2015-09-20
Better ordering of pattern variants in toolbar menu

e2tha-e:
Ref issue #153 
Just a change in the order of execution. Please review, comment, and merge.


e2tha-e:
Closing because I need to pr against dev

Issue #156, opened 2015-10-02
Need recursive partial inclusion


e2tha-e:
Bug: Since partial inclusion is currently iterative, if a Mustache file includes a partial that diveSync will iterate on later because it is sorted higher alphanumerically, Pattern Lab will fail.

To replicate:
* Create `source/_patterns/02-organisms/00-global/02-test.mustache`
* Enter some dummy content: `<h1>Foo</h1>`
* Include that partial anywhere in `source/_patterns/02-organisms/00-global/00-header.mustache`
* `{{> organisms-test }}`
* View the Pattern Lab in a browser - completely broken

Fix: need recursive inclusion of partials

bmuenzenmeyer:
This will auto-close when it reaches `master`

PR #157, opened 2015-10-02
Recursive partial include

e2tha-e:
Ref issue #156 
I think this is fairly well commented, but here's a general overview:

diveSync iterates through all pattern files once to populate the patternlab.patterns array. This is necessary because a recursive inclusion of partials requires that all partial patterns exist in that array. Prior to this pr, any attempt to include partials occurring later in alphanumeric order would fail and break patternlab, because the patternlab.patterns data that the include operation was looking for was non-existent.

Therefore, another diveSync operation is required to do the actual recursion. This second diveSync successfully populates the extendedTemplate properties because the template properties were previously populated for all partials in the previous diveSync operation.

Please review, comment, and merge.


bmuenzenmeyer:
Hi @e2tha-e this sounds like a welcome addition. I will certainly check it out - as it addresses a long-standing quirk/known issue.

e2tha-e:
@bmuenzenmeyer Please note the additional commits address some ordering and text-rendering concerns in the navbar.

* Commits 9441cec and 9e1fc5d keep variants grouped with their respective parents by maintaining their order in the patternlab.patterns array as per the diveSync.
* Commit 34793f5 allows patterns to include underscore-prefixed patterns, but does not display them in the UI. Previously, patternlab-node treated underscore-prefixed patterns as if they didn't exist.
* Commit 0109d6e allows for naming subdirectories without numeric prefixes and with any number of hyphens.

These code changes were mostly within parts I already edited, so I felt they shouldn't be in another pr.

PR #179, opened 2015-11-08
Remove extra/unnecessary word

gobrendan:
It's a pleasure to contribute, even in a small way :D

I'm a big fan of Pattern Lab -- been using it on projects for two years.


bmuenzenmeyer:
@gobrendan thanks!

PR #197, opened 2015-12-01
Update README.md

RichardBray:
I had to run the command I added for it to work


bmuenzenmeyer:
Hi Richard

This is good to know but it's platform specific. I think it would be better to mention that users may need administrative privileges, whether that be by running `sudo` or opening a command prompt as administrator. If you could add a sentence to that effect after these steps instead of what is currently here, I would accept this PR. Otherwise I've added it to my to-do list.

bmuenzenmeyer:
Closed with https://github.com/pattern-lab/patternlab-node/commit/ed4f8dd63e2a5989cf04169e2e1dd1d377f55549

PR #205, opened 2015-12-08
Changed grunt test script to gulp test script

dirkpennings:


bmuenzenmeyer:
Sorry for the confusion!
As of right now, this is only used by Travis CI to run the unit tests, which uses grunt as it's only current configuration. You can see that in the [travis file](https://github.com/pattern-lab/patternlab-node/blob/master/.travis.yml). That's why you still see this here even in the gulp side. 

Thanks for caring enough to contribute back to Pattern Lab Node. I'd love to know how you are using the tool and see more from you. Thanks again for looking out. Brian

dirkpennings:
Ok, thanks for explaining. I did try `gulp travis` in the command line and it seemed to work (don't know Travis that well), therefor I made the pull request.

bmuenzenmeyer:
If I recall, you can call gulp nodeunit to run unit tests. These are only
needed for those developing/hacking on patternlsb instead of just using it.

Issue #221, opened 2016-01-04
Become `require()`-able by allowing to become a dependency


EvanLovely:
It would be great to be able to use Pattern Lab Node as a dependency in another project (like a CMS theme build for a client - my use case). I would like my repo to contain the source code for the Pattern Lab Patterns, Data, Config, and Header/Footer template files, but not the compiled source – kinda like how we commit our `*.scss` but not our `*.css`. To do this we'd need to be able to do this at the top of a Gulp file:

```js
var pl = require('patternlab-node')({
  "paths": {
    "source": {
      "root": "./source/",
      "patterns": "./source/_patterns/",
      "data": "./source/_data/"
    }
});

gulp.task('pl', function() {
  pl.build();
});
```

Basically pass in the config object (could just read a `config.json` in our repo and pass it in too) and then run the build command. 

Is there any interest in the project moving in this direction? I could see the gulp and grunt "starter kits" you already provide getting ported to a yeoman generator for people who want all the stuff out of the box. 

Thanks so much for your work and interest in the Pattern Lab world!! :beers:

bmuenzenmeyer:
hey @EvanLovely 

I love the idea and the example syntax. For what it's worth [this PR just landed in dev](https://github.com/pattern-lab/patternlab-node/pull/216) and I have plans to get it into [1.1.0](https://github.com/pattern-lab/patternlab-node/milestones/v1.1.0) along with some example uses. 

You'll also see that @geoffp included an example usage. It's not ready, but it's coming!

EvanLovely:
OK @bmuenzenmeyer - I've got the basics up in #223 !

bmuenzenmeyer:
Did the last commit of #223 address this issue in its entirety?

EvanLovely:
It's a huge step forward, but I think for this issue to be complete, we'd need it documented and any other steps ironed out. The PR #223 laid most of the groundwork (hence the "Basic" part of the title).

bmuenzenmeyer:
I am freed up, and will start testing and documenting.

bmuenzenmeyer:
Documenting this has moved to https://github.com/pattern-lab/patternlab-node/wiki/Running-Pattern-Lab-Node-as-an-NPM-Dependency

Finishing https://github.com/pattern-lab/patternlab-node/pull/237 has helped simplify the styleguide steps quite a bit. When I pick this up again, I should be able to push through the remaining tasks and write up grunt docs too.

[...]

Issue #220, opened 2016-01-01
Support Configurable Pattern Lab Paths/Filesystem


bmuenzenmeyer:
This is the wrap up from [Configurable paths galore](https://github.com/pattern-lab/patternlab-node/pull/216)

## TODO

- [x] Initial work
- [x] gruntfile @bmuenzenmeyer
- [x] [dev branch: Add copy commands for public/styleguide assets](https://github.com/pattern-lab/patternlab-node/issues/229)
       - (short term fix - until https://github.com/pattern-lab/patternlab-node/issues/133 lands)
- [x] [dev branch: Config Paths only work with `./` prefix](https://github.com/pattern-lab/patternlab-node/issues/222)
- [x] more testing - grunt
   - **absolute**
   - [x] Move source.root directory to an arbitrary absolute path
   - [x] Move source.patterns directory to an arbitrary absolute path
   - [x] Move source.styleguide directory to an arbitrary absolute path   
   - [x] Move source.fonts directory to an arbitrary absolute path   
   - [x] Move public.root directory to an arbitrary absolute path
   - **relative**   
   - [x] Move source.root directory to an arbitrary relative path
   - [x] Move source.patterns directory to an arbitrary relative path
   - [x] Move source.styleguide directory to an arbitrary relative path
   - [x] Move source.fonts directory to an arbitrary absolute path   
   - [x] Move public.root directory to an arbitrary relative path
- [x] more testing - gulp
   - **absolute**
   - [x] Move source.root directory to an arbitrary absolute path
   - [x] Move source.patterns directory to an arbitrary absolute path
   - [x] Move source.styleguide directory to an arbitrary absolute path   
   - [x] Move source.fonts directory to an arbitrary absolute path   
   - [x] Move public.root directory to an arbitrary absolute path
   - **relative**   
   - [x] Move source.root directory to an arbitrary relative path
   - [x] Move source.patterns directory to an arbitrary relative path
   - [x] Move source.styleguide directory to an arbitrary relative path
   - [x] Move source.fonts directory to an arbitrary absolute path   
   - [x] Move public.root directory to an arbitrary relative path
- [x] final check with Geoff
- [x] final check with Evan

bmuenzenmeyer:
Made some progress on this tonight.

geoffp:
:+1:

bmuenzenmeyer:
@geoffp any further testing you can do on this is appreciated.
some thoughts, that I will likely do too:

[...]

PR #223, opened 2016-01-05
Basics of being dependent and passing in config

EvanLovely:
OK, so here's the absolute basics of being able to `var pl = require('patternlab-node')(config);`. The big problem that remains is that most of the stuff that is ran from `builder/patternlab.js` (the `require`-ed file) is running from the working directory of `node_modules/patternlab-node/`, so this PR doesn't complete #221 but simply moves towards it and allows it to work if you know what you're doing. The way I'm getting around this (for now) is with this:

```js
var config = fs.readFileSync('./config--pattern-lab.json');
var pl = require('patternlab-node')(config);
// get directory
var plnDir = path.dirname(require.resolve('patternlab-node/package.json'));
// save place
var originCwd = process.cwd();
// change to pl directory
process.chdir(plnDir);
// build it
pl.build(true);
// come back
process.chdir(originCwd);
```

This PR also creates the distinction between dependencies and devDependencies – if this is a required dependency in other projects, only the dependencies get installed; otherwise when `npm install` is ran here, both get installed. 

I expect this PR to be easy to merge and should not disrupt other workflows or uses and hopefully we can build upon this approach by making `builder/patternlab.js` run in the correct working directory (I believe it's a distinction between `process.cwd()` and `__dirname`, but I'd need to look closer to really know).

🍻

### DIFF
--- a/builder/patternlab.js
+++ b/builder/patternlab.js
@@ -25,7 +25,7 @@ var patternlab_engine = function (config) {
   patternlab = {};
 
   patternlab.package = fs.readJSONSync('./package.json');
-  patternlab.config = config || fs.readJSONSync('./config.json');
+  patternlab.config = config || fs.readJSONSync(path.resolve(__dirname, '../config.json'));
 
   var paths = patternlab.config.paths;
 
--- a/builder/patternlab.js
+++ b/builder/patternlab.js
@@ -8,7 +8,7 @@
  *
  */
 
-var patternlab_engine = function () {
+var patternlab_engine = function (config) {
   'use strict';
 
   var path = require('path'),
@@ -25,7 +25,7 @@ var patternlab_engine = function () {
   patternlab = {};
 
   patternlab.package = fs.readJSONSync('./package.json');
-  patternlab.config = fs.readJSONSync('./config.json');
+  patternlab.config = config || fs.readJSONSync('./config.json');
 
   var paths = patternlab.config.paths;
 
--- a/package.gulp.json
+++ b/package.gulp.json
@@ -2,21 +2,24 @@
   "name": "patternlab-node",
   "description": "Pattern Lab is a collection of tools to help you create atomic design systems. This is the node command line interface (CLI).",
   "version": "1.0.1",
-  "devDependencies": {
-    "browser-sync": "^2.10.0",
+  "main": "./builder/patternlab.js",
+  "dependencies": {
     "del": "^2.0.2",
     "diveSync": "^0.3.0",
     "fs-extra": "^0.26.2",
     "glob": "^6.0.1",
+    "html-entities": "^1.2.0",
+    "mustache": "^2.2.0"
+  },
+  "devDependencies": {
+    "browser-sync": "^2.10.0",
     "gulp": "^3.9.0",
     "gulp-connect": "^2.2.0",
     "gulp-copy": "0.0.2",
     "gulp-header": "^1.7.1",
     "gulp-load": "^0.1.1",
     "gulp-nodeunit": "0.0.5",
-    "gulp-strip-banner": "0.0.2",
-    "html-entities": "^1.2.0",
-    "mustache": "^2.2.0"
+    "gulp-strip-banner": "0.0.2"
   },
   "keywords": [
     "Pattern Lab",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "version": "1.0.1",
   "main": "./builder/patternlab.js",
   "dependencies": {
-    "bs-html-injector": "^3.0.0",
     "diveSync": "^0.3.0",
     "fs-extra": "^0.26.2",
     "glob": "^6.0.1",
@@ -13,6 +12,7 @@
     "mustache": "^2.2.0"
   },
   "devDependencies": {
+    "bs-html-injector": "^3.0.0",
     "grunt": "~0.4.5",
     "grunt-browser-sync": "^2.2.0",
     "grunt-contrib-concat": "^0.5.1",
--- a/package.json
+++ b/package.json
@@ -2,20 +2,23 @@
   "name": "patternlab-node",
   "description": "Pattern Lab is a collection of tools to help you create atomic design systems. This is the node command line interface (CLI).",
   "version": "1.0.1",
-  "devDependencies": {
+  "main": "./builder/patternlab.js",
+  "dependencies": {
     "bs-html-injector": "^3.0.0",
     "diveSync": "^0.3.0",
     "fs-extra": "^0.26.2",
     "glob": "^6.0.1",
+    "html-entities": "^1.2.0",
+    "matchdep": "^1.0.0",
+    "mustache": "^2.2.0"
+  },
+  "devDependencies": {
     "grunt": "~0.4.5",
     "grunt-browser-sync": "^2.2.0",
     "grunt-contrib-concat": "^0.5.1",
     "grunt-contrib-copy": "^0.8.2",
     "grunt-contrib-nodeunit": "^0.4.1",
-    "grunt-contrib-watch": "^0.6.1",
-    "html-entities": "^1.2.0",
-    "matchdep": "^1.0.0",
-    "mustache": "^2.2.0"
+    "grunt-contrib-watch": "^0.6.1"
   },
   "keywords": [
     "Pattern Lab",


bmuenzenmeyer:
@EvanLovely you are an open source juggernaut! Thanks for this work - I am feeling a bit overwhelmed by the support and interest of late, so I might be slow on reviewing everything. Please put up with me :hourglass_flowing_sand:

EvanLovely:
Thanks! No worries. Hopefully I can help more soon as we use PatternLab pretty heavily at work.

bmuenzenmeyer:
Hi @EvanLovely forgive my ignorance on this following issue...

If this is accepted, creating the distinction between devDependencies and dependencies, won't a user pulling down patternlab node via `npm install patternlab-node` and running "vanilla" only receive the dependencies? I feel this could be detrimental without proper documentation, as people simply running `npm install patternlab-node` may not realize they will not get the entire harness.... whereas users intending to use patternlab node as a dependency will definitely have upstream task harnesses/etc.

... thinking and googling ensues ...

This is just now starting to click more for me! Ok. So really the "vanilla" documentation should be moe clear (this is more for my benefit than yours, FYI don't be offended :blush: ):

---
### Download
You may get Pattern Lab a couple ways, depending on your intended use:

**Stand-Alone Installation**
* Download the [latest release of patternlab-node](https://github.com/pattern-lab/patternlab-node/releases/latest) from Github
* or clone the git repository [https://github.com/pattern-lab/patternlab-node.git](https://github.com/pattern-lab/patternlab-node.git)
* From the command line at your extracted file location, run `npm install` 
* **NOTE** Doing this will install grunt (or gulp if you follow the instructions later in this README) along with the helper libraries necessary to shuffle assets and serve the Pattern Lab website / styleguide.

**Pattern Lab as a Dependency**
* run `npm install patternlab-node`
* **NOTE** Doing this will omit grunt, gulp and their respective helper libraries, assuming you are handling a lot of asset shuffling upstream and only using the `./builder/patternlab.js` API.

----

Please let me know if I have this right. If I do, I'll accept this and get the `README` updated to reflect the above after some testing.

cc @geoffp this might be of interest to you too.

EvanLovely:
I think that putting out documentation on how to use this as a dependency is too early. Due to how the paths are set up in `builder/patternlab.js` one needs to change the working directory to `node_modules/patternlab-node/` for the builder to work (even with custom paths). This more or less lays the ground work for this repo to be used as a dependency. 

I also think that a Yeoman Generator would be the way to go for scaffolding out a "Stand-Alone Installation" – we could even ask questions like if they want Sass or not, Grunt or Gulp, etc and then create a different file structure based on those questions. 

Also, a good tidbit you may find useful for testing is the ability to [supply different arguments to `npm install`](https://docs.npmjs.com/cli/install); so if you wanted to see what a `npm install patternlab-node` would be like after this PR merges, you can run:

    npm install git@github.com:EvanLovely/patternlab-node.git#feature/enable-as-dependency

or a shorthand for GitHub URLs:

    npm install EvanLovely/patternlab-node#feature/enable-as-dependency

Doing so and comparing to `npm install patternlab-node` I can see my version installs the stuff in `dependencies` into `node_modules/patternlab-node/node_modules/` whereas the regular installs nothing into it (b/c all of them are `devDependencies`). In either case if `npm install` is ran in the `patternlab-node/` directory the same set of dependencies are installed: all of them.

So basically what I'm saying is that I don't think anything needs to get added to the README: all dependencies will get installed with a `npm install` as documented after a download and the process to run this repo as a dependency is too cumbersome for mass consumption due to working directory challenges; however this will set the foundation for that work to be done – work I look forward to helping with.

bmuenzenmeyer:
> I think that putting out documentation on how to use this as a dependency is too early. Due to how the paths are set up in builder/patternlab.js one needs to change the working directory to node_modules/patternlab-node/

Okay I finally get this. Your example code was useful, it wasn't clicking the first time through my head. Very happy to have you on board to help with thinking through this problem!

> I also think that a Yeoman Generator would be the way to go for scaffolding out a "Stand-Alone Installation" – we could even ask questions like if they want Sass or not, Grunt or Gulp, etc and then create a different file structure based on those questions.

I like the idea in general, but at least at this time I want to keep Yeoman outside the scope of this repository. I think that could change as it matures more (or perhaps others focus on this), but right now my focus is this dependency work, pattern engines, and making things more modular/extensible...I believe others have already wrapped the project and are essentially doing this work, but I don't recall who or where at this time. I've wondered if my perspective toward Yeoman is skewed because I don't go through the motions/pain of initializing projects with the same frequency as those in client services.

Thanks for the tips on testing. I've tested PRs locally for a while, but never tried `npm installs` - so that's pretty cool. Now that I understand the state of this issue more, I'll try to get it merged post-haste.

geoffp:
It seems like the struggle here is against direct references to `config.json`. Here's my quick search:

```
builder/lineage_hunter.js:20:29:            var config = require('../config.json');
builder/list_item_hunter.js:20:26:          config = require('../config.json'),
builder/object_factory.js:17:35:            var config = fs.readJSONSync('./config.json');
builder/pattern_assembler.js:21:37:         var config = fs.readJSONSync('./config.json');
builder/patternlab.js:27:42:                patternlab.config = fs.readJSONSync('./config.json');
test/engine_handlebars_tests.js:23:27:      config: require('../config.json'),
test/pattern_assembler_tests.js:290:43:     patternlab.config = fs.readJSONSync('./config.json');
test/pattern_assembler_tests.js:615:43:     patternlab.config = fs.readJSONSync('./config.json');
test/pattern_assembler_tests.js:707:43:     patternlab.config = fs.readJSONSync('./config.json');
```

The first one is completely unnecessary, the next two just need the debug flag, but don't have access to the main patternlab object, which is where the config data lives. The rest are unit tests that may not ever have access to the real patternlab object -- that's a challenge.

Maybe we can cheat and create a global variable (not even sure how that works in Node, honestly) for the debug flag. And maybe we can just say that the unit tests only work in a standalone PL -- I personally only run them in a standalone copy.

geoffp:
Also, I agree with @bmuenzenmeyer on Yeoman -- it's a fine tool where it works, and I'm not against code generation or scaffolding in general. But I think we're trying to aggressively reduce the number of moving parts and external dependencies needed to get up and running with PL to tear down barriers to use.

bmuenzenmeyer:
> Maybe we can cheat and create a global variable (not even sure how that works in Node, honestly) for the debug flag

I've started reading more of the node documentation to determine how best to do this, but yet I think it's possible to place this flag somewhere more intelligently. 

https://nodejs.org/api/modules.html
https://nodejs.org/api/globals.html

geoffp:
It might be nice to be able to say something like:
```sh
grunt serve --debug
gulp serve --debug
```
rather than editing the config.

bmuenzenmeyer:
Agreed

EvanLovely:
OK, lots to respond to here, so I'm going to do this in parts; first off @geoffp I'm not seeing those hard coded references to `config.json` you mention 🤔 – are you on `dev`? I checked in `master` too.

Here's the first few you mentioned and where I'm looking:

- https://github.com/pattern-lab/patternlab-node/blob/dev/builder/lineage_hunter.js#L20
- https://github.com/pattern-lab/patternlab-node/blob/dev/builder/list_item_hunter.js#L20
- https://github.com/pattern-lab/patternlab-node/blob/dev/builder/builder/object_factory.js#L17

The only two references I can find to `config.json` are [`test/pattern_assembler_tests.js#L212`](https://github.com/pattern-lab/patternlab-node/blob/dev/test/pattern_assembler_tests.js#L212) (which I think is fine to leave) and the change I'm proposing in `builder/patternlab.js`.

EvanLovely:
Alright, next up: I've just [added a new commit](https://github.com/EvanLovely/patternlab-node/commit/e73b21d753193b61ab99d831dff20051e41cf76c?diff=unified) to this PR. When references to `config.json` are resolved using, the default is to resolve from where the command was ran from, the current working directory, which this project has made the understandable assumption will be the root of this repo. However, if it's ran from somewhere else, say two directories up, then references to `config.json` are resolved from there. 

This PR enables people to pass in a `config` object to use instead of `config.json`, but we still would like it to get read as a default, so my last commit utilizes the [`__dirname` variable](https://nodejs.org/api/globals.html#globals_dirname) (The name of the directory that the currently executing script resides in.) and a `path.resolve()` to get to the config: basically saying: start from directory where this file resides, and go find `../config.json` – instead of: find `config.json` in this current working directory.

```diff
-  patternlab.config = config || fs.readJSONSync('./config.json');
+  patternlab.config = config || fs.readJSONSync(path.resolve(__dirname, '../config.json'));
```

I believe this pattern could be applied to other places and that would solve having change the working directory to the root of this repo before executing.

geoffp:
@EvanLovely you're right, I think many of those are probably just in the pattern-engines branch, which is where I'm working at the moment. Can you address the ones in the pattern_assembler tests?

This new approach makes me way less nervous than changing the working directory.

EvanLovely:
Good to hear! I think the reference to `./config.json` in [`test/pattern_assembler_tests.js#L212`](https://github.com/pattern-lab/patternlab-node/blob/dev/test/pattern_assembler_tests.js#L212) is fine to leave as that's a test and those will definitely be ran from the root of this repo and not from a parent project.

Agreed on not like the approach on changing the working directory – that's why I suggested not documenting that method for others to follow.

bmuenzenmeyer:
I checked this out and merged after a quick test.

@geoffp  and @EvanLovely - from your perspectives, what's left to address on this PR or https://github.com/pattern-lab/patternlab-node/issues/220?

I am aware of #229 and will be taking that one

geoffp:
It would be nice but not necessary to have a global variable for the debug flag.

I'd like to try replacing my copy-the-config solution with this one today, if I can get to it, since I have a real-world NPMified configuration up and running. That should give us one more vote of confidence.

geoffp:
It works for me! My gulpfile is much simpler without the fussy config file copy step, and everything seems to work. I guess all we need now is some documentation.

@bmuenzenmeyer, feel free to assign me any work you need to get 1.1.0 out the door. I'm going to work on some of our internal toolchain stuff until I hear back.

bmuenzenmeyer:
Great!

@geoffp thanks for the offer of help with 1.1.0 - I definitely appreciate and welcome it.

Could toy please take the README content stubbed out in https://github.com/pattern-lab/patternlab-node/pull/223#issuecomment-169905848 as a base and see if anything else is necessary?

geoffp:
On it.

geoffp:
Okay -- I've gone through a "fresh install" from scratch and tried to piece together a fresh require()able copy of patlab, documenting each step I had to take.

This is going to bring into focus why it's not quite yet a supported feature, if you know what I mean. :grinning: But it should also tell us what we need to streamline to make it a better supported configuration. Full markdown coming up next. I have the .md stored on my machine as well, so I can hand it off properly if/when you want to publish these instructions.

bmuenzenmeyer:
:+1: @geoffp thanks so much!

geoffp:
@EvanLovely, would you review this as well, please?

#### Installing and running Pattern Lab as an NPM dependency

**Note**: support for this configuration is still young, and is still being actively worked on. Expect a few bumps in the road and bear with us as we polish this up.
**Prerequisites**: Node/NPM, and a project with gulp and/or grunt installed. Commands are given in bash.

1. **Install Pattern Lab**: Go to your project root and run `npm install --save-dev patternlab-node`
1. **Copy the default config**: you can call it whatever you want, but in this example, we do `cp node_modules/patternlab-node/config.json patternlab-config.json`
1. **Copy the default source**: you need a basic copy of the patternlab source files to get started. Do `cp -r node_modules/patternlab-node/source .`
1. **Copy the Gulpfile or Gruntfile**: Either `cp node_modules/patternlab-node/gulpfile.js .` or `cp node_modules/patternlab-node/Gruntfile.js .`
1. **Edit your config**: 
 1. In the `path` section at the top, tell Pattern Lab where to build its static HTML site; the default is `[project root]/public`.
 1. Add a `styleguide` entry to paths.source to tell gulp where to copy styleguide assets from. By default this should be `./node_modules/patternlab-node/public/styleguide`
1. **(Gulp only) Edit your gulpfile**:
 1. Install NPM dependencies needed for the gulpfile: `npm install --save-dev del gulp-strip-banner gulp-header gulp-nodeunit gulp-load browser-sync`
 1. Edit the `paths()` function to point to `patternlab-config.json` or whatever you chose to call your custom config file.
 1. Comment out this line, currently at or near line 29: `gulp.loadTasks(__dirname+'/builder/patternlab_gulp.js');` We need our own version of the `patternlab` task, not the default that's included.
 1. Create a custom `patternlab` task that loads your project's pattern lab config and passes it in. Something like:

    ```javascript
    // The config and the Pattern Lab itself
    var config = require('./patternlab-config.json'),
        pl = require('patternlab-node')(config);

    gulp.task('patternlab', ['prelab'], function (done) {
      pl.build(true);
      done();
    });
    ```

 1. Create a task to copy over the styleguide assets. Something like:

    ```javascript
    // Styleguide assets copy from node module to public dir
    gulp.task('cp:styleguide', ['clean'], function () {
      return gulp.src(path.resolve(paths().source.styleguide, '**/*'))
        .pipe(gulp.dest(paths().public.styleguide))
        .on('end', function () {
          console.log('Done copying styleguide stuff.');
        });
    });
    ```

 1. Add your styleguide copy task to the `prelab` task, as in:

    ```javascript
    gulp.task('prelab', ['clean', 'assets', 'cp:styleguide']);
    ```

 1. Run it and see if it works!

    ```bash
    gulp serve
    ```

1. **(Grunt only) Edit your Gruntfile**:
 1. **TODO**

EvanLovely:
Hey, sorry for my lack of responses — was on a birthday ski vacation :) I'll check into this very soon; thanks!!

Issue #227, opened 2016-01-07
gulpfile: style.css not copied after clean install & browser-sync watching wrong css directory


robinsonaaron:
After running "gulp serve" from a clean install, the style.css file is not copied to the public directory. Also, browser-sync is watching the public css directory when it should be watching the source directory.

bmuenzenmeyer:
This will close automatically when it reaches `master`

PR #228, opened 2016-01-07
Fix gulpfile issues with copying and watching css files.

robinsonaaron:
Fixes issue #227 "gulpfile: style.css not copied after clean install & browser-sync watching wrong css directory".

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -101,7 +101,7 @@ gulp.task('connect', ['lab'], function(){
       baseDir: paths().public.root
     }
   });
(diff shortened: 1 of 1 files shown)


geoffp:
I think that was my oversight. Thanks for this fix!

bmuenzenmeyer:
Yes I think this looks good. I will be checking the grunt version.
Good eyes. Think it was introduced here  https://github.com/pattern-lab/patternlab-node/commit/9dcfc63b99964757bcd9d085d39d87fd2e715e63

PR #230, opened 2016-01-08
Root out a few more hard-coded paths

geoffp:
That, and kill more `file.substring(2)`s that were a potential source of errors.

PR #233, opened 2016-01-12
Pattern engines ongoing work, round 5 (I think)

geoffp:
Happy New Year! I feel this work is getting very close to stable at this point. I mean, it passes the tests, what more do you want?

Actually, the latest merge from the dev branch was a big one, and nicely exposed some Mustache-specific code still in the lineage hunter. I factored it out, and then moved it into the engines -- only to find that it was duplicating a function already present in the engines! The "old" version extracted from the lineage hunter was passing more tests than the "new" one, so at least for the Mustache engine, the old one wins.

More notable stuff:
* More file-based unit tests for Handlebars -- these test that the JSON data environment for Handlebars partials is correct.
* One bug I had after the merge was that the processing of style modifiers and pattern parameters was breaking Handlebars in certain cases. There was a big chunk of `processPatternRecursive()` that was doing only this, [so this was factored out](https://github.com/geoffp/patternlab-node/commit/724396735eeb20c21ddc582094cc7d1786959618) into an `expandPartials()` function (in the same file) that is only executed if the engine of the currently in-process pattern has an `expandPartials` flag set to `true`. I broadly like the extraction of that function, because it makes `processPatternRecursive()` shorter and easier to understand, however:
 * Even though this makes everything work for now, I'm not happy with it -- I think the cleanest solution is different, and I don't want to have to put a flag into engines that is rarely used and may be confusing to authors.
 * I think we've talked about this before, but it does seem to me like the `expandPartials()` functionality really belongs in the Mustache engine, since style modifiers and pattern parameters amount to Mustache-specific syntax extensions. They don't seem necessary in Handlebars, anyway, because Handlebars is already rich with argument-passing features. Doesn't it seem kind of fun that PL has its own dialect of Mustache? Pattern-lab flavored Mustache! Milk Mustache? There has to be a joke in there somewhere.

As always, I look forward to your thoughts!

### TODO
- [ ] don't execute Handlebars unit tests unless handlebars is installed via npm
- [ ] await @bmuenzenmeyer's feedback on moving params/stylemodifiers into Mustache plugin


bmuenzenmeyer:
@geoffp dumb question incoming

This completely ignores any work we are doing in `dev` with configurable paths, yes?

geoffp:
No, not at all -- this is tracking `dev`, so it's got any configurable path work that's merged to `dev`, which is everything except for [this, which looks like is landing soon](https://github.com/pattern-lab/patternlab-node/pull/230).

Totally fine of course if you want to focus on 1.1, I'm just keeping up to date with dev here and continuing to stabilize, fix and add tests as I test the Handlebars engine against our existing library.

Can I help the anxiety at all?

bmuenzenmeyer:
All yours @geoffp !

PR #235, opened 2016-01-14
Fixed typo in the package.gulp.json

simonknittel:


bmuenzenmeyer:
Hi @simonknittel 

This is confusing on my part, and should probably just be removed. Please see https://github.com/pattern-lab/patternlab-node/pull/205 for a longer explanation, but really this config entry is just for TravisCI and always runs using grunt

PR #237, opened 2016-01-28
Created /core/ and moved styleguide

bmuenzenmeyer:
I also removed scss files and config entries from gulp and grunt as they will eventually go away anyways.

### Issues
- [x] Does relying on grunt/gulp copy tasks makes this a dependency? Right now I have any pattern changes causing the styleguide to be rebuilt and rewritten to  /core/ so the watch tasks can grab that and then move everything over. Also I just realized that grunt and gulp perform patternlab and copy tasks in different orders. Either way, this all feels a bit rushed and I think I want more thought on it.
- [x] Add HUGE thank you to Geoff for configurable paths work


builder/patternlab.js
```diff
@@ -191,7 +191,7 @@ var patternlab_engine = function (config) {
     //build the styleguide
     var styleguideTemplate = fs.readFileSync(path.resolve(paths.source.patternlabFiles, 'styleguide.mustache'), 'utf8'),
     styleguideHtml = pattern_assembler.renderPattern(styleguideTemplate, {partials: styleguidePatterns});
-    fs.outputFileSync(path.resolve(paths.public.styleguide, 'html/styleguide.html'), styleguideHtml);
+    fs.outputFileSync(path.resolve(paths.source.styleguide, 'html/styleguide.html'), styleguideHtml);
```
bmuenzenmeyer: Doing this is what makes grunt/gulp copies dependencies right now....

PR #251, opened 2016-02-18
pattern_assembler regex cleanup, extensibility

e2tha-e:
@bmuenzenmeyer Please review. This is precursor work to what I hope will be a fix for issue #250

Issue #250, opened 2016-02-17
Patterns with Pattern Parameters Cannot Be Overwritten Downstream


e2tha-e:
To replicate:
- Edit `source/_patterns/04-pages/00-homepage.mustache`
- Replace the content with this mustache code: `{{> templates-homepage(emergency: true) }}`
- This will render correctly in the Pattern Lab UI.
- Edit `source/_patterns/03-templates/00-homepage.mustache`
- Replace `{{> molecules-alert }}` with `{{> molecules-alert(alertClass: 'error') }}`
- This causes the emergency block to disappear instead of appearing with the correct alert class.

bmuenzenmeyer:
To my knowledge, this is the expected behavior since pattern parameters were released - since encountering a pattern parameter causes immediate consumption of the data effectively closing off downstream changes. 

Does having a file like `.source/_patterns/04-pages/00-homepage~emergency.json` work for you?

e2tha-e:
I remember doing recursive parameter submission all the time in Pattern Lab PHP, but wanted to be absolutely sure the latest version does it, and v1.0.0 definitely does it. Making the the exact code changes as in my first comment gets the alert class to show correctly.

Tilde-appended variants aren't the best option if you want to view components up and down the stream because you'll repeat definitions in multiple .json files, which goes against the DRY principle.

In any case, the emergency block shouldn't just disappear because of a downstream parameter submission.

bmuenzenmeyer:
Hey @e2tha-e thanks for checking. I think I have a proposed solution that I might hack out and try as a branch. Welcome to suggestions for anyone too

theorise:
I just wanted to say I came across this just yesterday and it is certainly something that would be great to have. 

I may be abusing pattern parameters, but I am passing things such as data-attributes as JS hooks for components. In my use case: 

We have a reusable icon atom, inside a reusable off-canvas organism. The icon acts as the close button:

`{{> atoms-icon:icon--cross(iconDataAttributes:'data-off-canvas-toggle="toggle-target"') }`

The molecule is included in certain templates, but it needs to be included as the following:

`{{> organisms-off-canvas:off-canvas_small|off-canvas_right(offCanvasSelector:'toggle-target') }}`

bmuenzenmeyer:
Morning

The specifics of this issue seemed acceptable to me for a long time since only people using the more advanced features under certain scenarios would encounter this problem, and really one is left with many options as workarounds.

[...]

PR #253, opened 2016-02-18
Regex cleanup

e2tha-e:
Referencing comment in #251
From @bmuenzenmeyer : 
> Hi @e2tha-e
> 
> Thanks for this contribution - I am nervous though of its eventual collisions with the pattern_engine branch which already refactored a ton of regexes.
>
> @geoffp could you help weigh in?

### DIFF
--- a/builder/pattern_assembler.js
+++ b/builder/pattern_assembler.js
@@ -44,7 +44,7 @@
     // returns any patterns that match {{> value:mod }} or {{> value:mod(foo:"bar") }} within the pattern
     function findPartialsWithStyleModifiers(pattern){
       // the regular expression broken down:
-      // the opening mustache include tag
+      // opening mustache include tag
       var regexStr = '{{>';
       // an optional single space
       regexStr += '( )?';
@@ -64,7 +64,7 @@
       regexStr += '(\\([^\\)]*\\))?';
       // an optional single space
       regexStr += '( )?';
-      // the closing mustache tag
+      // closing mustache tag
       regexStr += '}}';
 
       var regex = new RegExp(regexStr, 'g');
@@ -74,7 +74,7 @@
     // returns any patterns that match {{> value(foo:"bar") }} or {{> value:mod(foo:"bar") }} within the pattern
     function findPartialsWithPatternParameters(pattern){
       // the regular expression broken down:
-      // the opening mustache include tag
+      // opening mustache include tag
       var regexStr = '{{>';
       // an optional single space
       regexStr += '( )?';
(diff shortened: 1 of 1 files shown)


geoffp:
For sure, I'll hopefully get time to look at his today.

e2tha-e:
@bmuenzenmeyer @geoffp 
Hopefully we can converge to a point where regexes are both cleaned up and well documented. I haven't looked through the pattern_engine branch, so pardon me if I'm redoing work that has already been done, or colliding in an irrecoverable way.
Here are some things I am trying to accomplish in this pr:

- Allow partial-finding functions to parse strings as well as pattern objects.
- Simplify regexes in the following ways:
  - `( )` === `([ ])` - The square brackets are unnecessary and even excluded on the `findListItems` function.
  - `\w` === `[A-Za-z0-9_]`
  - `(?:` - the 'g' global switch for `String.prototype.match` ignores group capture, so the regex syntax to not capture does nothing but add noise.
  -  `(| )` === `([ ])?` === `( )?` - Use one and use it consistently.
- Strengthen regexes in at least this way:
  - `(\([^\)]*\))` - The old regex searched for an opening parenthesis followed by any number of characters, followed by an optional space, terminating in double closing curly braces. The new regex ensures that parentheses enclose the parameter(s). Then, the regex continues to search for the optional space and double closing curly braces.

[Mozilla docs on RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp)

geoffp:
All worthy goals!

Yes, unfortunately, all these functions are Mustache-specific, and have moved to the Mustache engine in the pattern-engines branch. Would you mind re-evaluating this against the regexes in `builder/pattern_engines/engine_mustache.js` in that branch?

bmuenzenmeyer:
I'd go one step further and say that this PR should be closed. I do however love the idea of better documenting the specific regexes, and it looks like some improvements to the regex logic is due. 

I admit, however, that I don't remember how they are documented in the `pattern-engine` branch currently. 

I echo @geoffp in looking to that branch.

---

I am curious as to how you think this PR helps setup a solution for #250 
I have an idea as to how to make this work that seems low-tech but I haven't had time to try it yet.

e2tha-e:
@bmuenzenmeyer It sets up a solution for #250 because the partial finding functions would accept strings as well as the pattern object. And as I got into the weeds of making sure the regexes are matching correctly, I decided that breaking them apart and documenting them would help anyone else who needs to either augment or correct them in the future.

e2tha-e:
I'll definitely look at the pattern_engine branch and see if these changes are more appropriate there.

e2tha-e:
Closing this in favor of #256

PR #256, opened 2016-02-19
Regex and partial finding update on pattern-engines branch

e2tha-e:


e2tha-e:
@bmuenzenmeyer @geoffp Please review

bmuenzenmeyer:
@e2tha-e thanks for pivoting here - I appreciate the contribution and want it to see the light of day, and think this is the best route. pattern_engines are @geoffp's baby so I will let him rule in

geoffp:
This looks good! The structure is well thought out. My only reservation is that breaking up the regexes makes it harder to paste them into a regex tester, but honestly, it's such a nightmare to edit them when they get that long that I think it's a good trade-off.

geoffp:
@e2tha-e, if it breaks someday, now you have to fix. it. ;) Thanks for the good work!

e2tha-e:
@geoffp Haha. Right on. Already on it!

PR #257, opened 2016-02-19
update lineage_hunter - refactor

mateuspv:
**Target the `dev` branch!**
Summary of changes:
I remove Closures("(function () { ... })()") since node will not export any that are not implicit with `module.exports`. Also removed some nested functions.

Thanks!

### DIFF
--- a/builder/lineage_hunter.js
+++ b/builder/lineage_hunter.js
@@ -1,76 +1,90 @@
-/* 
- * patternlab-node - v1.1.1 - 2016 
- * 
- * Brian Muenzenmeyer, and the web community.
- * Licensed under the MIT license. 
- * 
- * Many thanks to Brad Frost and Dave Olsen for inspiration, encouragement, and advice. 
- *
- */
+/*
+* patternlab-node - v1.1.1 - 2016
+*
+* Brian Muenzenmeyer, and the web community.
+* Licensed under the MIT license.
+*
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
Hi @mateuspv 

it looks like this optimization could be applied to other builder/ files besides list_item_hunter.js, is that correct?

mateuspv:
Hi @bmuenzenmeyer, Yes. You can remove the "main code" from inside the closure:
(function() {
/* main code here */
})()

bmuenzenmeyer:
Thanks for the clarification. I'd rather see all the extraneous closures removed at once for consistencies sake. If you feel like tackling that, please do so. Else this might get prioritized a bit further down the pile.

geoffp:
I was just [researching this issue](http://stackoverflow.com/a/26083615/446564), and I agree that we should remove them all. 

@mateuspv, if you're up for doing all of them, that would be amazing, but if you don't want to, I could do it was part of a larger whitespace & indentation cleanup.

mateuspv:
@geoffp I'm sure when I will do it =/, so feel free to do..

bmuenzenmeyer:
I second rolling it into https://github.com/pattern-lab/patternlab-node/issues/264

bmuenzenmeyer:
Added as a todo item to https://github.com/pattern-lab/patternlab-node/issues/264.

Closing this.

Issue #252, opened 2016-02-18
browsersync injected both in the page and in the iframe


vlrprbttst:
do you know how to prevent this double injection? this is annoying because every time i save a file, browsersync fires but goes back to the top of the page and i have to scroll all the way down to the pattern i modified (when i'm the in ALL view).

i tried with [browsersync snipperOptions](https://www.browsersync.io/docs/options/#option-snippetOptions) but had no luck. I've been asking this on the [browsersync repo](https://github.com/BrowserSync/grunt-browser-sync/issues/129
) as well as here but had no chance as well.

do you have a solution for injecting the reload script __only__ in the `iframe` and not in patternlab overall `body`?

thanks! :)

bmuenzenmeyer:
Hi @vlrprbttst 

I agree this is annoying, and as a workaround will navigate to the pattern I am working on specifically if it happens to be "below the fold."

I'd love to fix this, and have devoted some nights to it in the past, but with no results. I will look at those two links you mention - but for now am going to close this as a duplicate of #159 

If you figure it out, let me know!

vlrprbttst:
of course!! only browserSync's and Patternlab creators could help at this point. because i've asked everywhere already :) i have a stack overflow question up as well: http://stackoverflow.com/questions/35293486/grunt-browser-sync-double-injection-of-the-reload-script

bmuenzenmeyer:
Neat - FYI on that question you state that it's Pattern Lab PHP you are using, not Pattern Lab Node.

vlrprbttst:
yeah, i first downloaded your project and noticed the issue, as an exercise I've rebuilt it with node and grunt. it's pretty similar to yours and we share the same exact problem so i guess this suits us both :)

geoffp:
I was just playing with this issue late last night. So late I can't
remember if I fixed it or not, but I did get CSS injection working
properly. I'll look at this in a few hours, hopefully.

vlrprbttst:
that would be awesome, thanks!

geoffp:
Anyone who's interested, please check out https://github.com/pattern-lab/patternlab-node/pull/260 and let me know if it works for you.

vlrprbttst:
@geoffp @bmuenzenmeyer what are the static paths of these?

`path.resolve(paths().public.root`
and
`path.resolve(paths().public.styleguide`

geoffp:
Those are (or should be) set in your patternlab config file. The relevant block looks like this:
```json
[...]

Issue #159, opened 2015-10-05
Optimize browsersync inject to avoid full-page refreshes


e2tha-e:
@bmuenzenmeyer @oscar-g
I can't seem to get browsersync to inject without a complete page refresh. If I load the pattern outside the iframe, then injection works fine. But inside the iframe, it causes the entire page to refresh. Does anyone know how to avoid the page refresh while still working within the Pattern Lab iframe?

bmuenzenmeyer:
Hey @e2tha-e browsersync is a bit clunky with patternlab due to the iframes. I've chalked this up as an acceptable quirk for now, citing the other big advantages it brings. I've mean to look more into this someday, unless a solution presents itself via PR.

e2tha-e:
@bmuenzenmeyer Do you think it would be ok to have Node listening on another port for the iframe? I don't know off-hand whether the same-origin policy will be fulfilled, and it seems like maybe too much overhead just for dom injection.

@geoffp mentioned he might have gotten injection working in #252. Eager to hear what it may be.

geoffp:
Anyone who's interested, please check out https://github.com/pattern-lab/patternlab-node/pull/260 and let me know if it works for you.

PR #260, opened 2016-02-22
Better browsersync

geoffp:
Addresses #252, #159 

**Summary of changes:**
This blacklists the UI chrome to exempt it from browsersync's snippet injection, meaning that the only thing it's glommed onto is the iframe itself and the stuff inside it. This should hopefully eliminate most of the spooky behavior that comes from BS, and in my testing, makes CSS injection and auto-page-refresh (upon editing pattern HTML) work correctly under both grunt & gulp.

**Bonus:** this also moves the browsersync badge to the lower right and chills it out some -- a request from my team that I think would be welcome to others.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -101,6 +101,10 @@ module.exports = function(grunt) {
 						ignoreInitial: true,
 						ignored: '*.html'
 					},
+                    snippetOptions: {
+                        // Ignore all HTML files within the templates folder
+                        blacklist: ['/index.html', '/']
+                    },
 					plugins: [
 						{
 							module: 'bs-html-injector',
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
@geoff I owe **you** a beer for this when you come my way :smile: :beers: 

will give it a quick review before merge.

Thanks friend!

geoffp:
Ah yeah, earnin' the :beers:

I had to read the browsersync code to figure this out -- the docs were not at all clear even on what the right option key was to use, whether it wanted a string, or array of strings, what the given path was relative to, yada yada. I think it's supposed to be a public URL fragment starting with a slash.

bmuenzenmeyer:
@geoffp thanks for this!

bramsmulders:
Guys, one quick note i'd like to point out from the top of my head. You NEED the refresh of the patternlab ui when adding or removing a pattern otherwise there will be no updated pattern menu. 

I thought of the solution mentioned above once, and i could not implement it in my own fork due to the problem outlined above.

geoffp:
Yeah, that's true -- I don't think I've ever been able to get a file-watcher to detect new files. Maybe once, but it was fragile. In fact, I think I once saw a bug report in Node.js itself about that.

That might be worth looking into again.

PR #262, opened 2016-02-24
Update destination of fonts to public.fonts

ryanatty:
Just a small thing I noticed while working with the gulpfile today, the fonts are being copied to public.images instead of public.fonts.


geoffp:
Good catch! Thanks @ryanatty!

PR #269, opened 2016-02-26
Add var declaration baseurl.

jeroenkampinga:
Added the variable declaration for "baseurl" in postmessage.js. Otherwise you receive a javascript error: "baseurl is not defined".

### DIFF
--- a/core/styleguide/js/postmessage.js
+++ b/core/styleguide/js/postmessage.js
@@ -10,7 +10,7 @@
 
 // alert the iframe parent that the pattern has loaded assuming this view was loaded in an iframe
 if (self != top) {
-	
+
 	// handle the options that could be sent to the parent window
 	//   - all get path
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
I'm not seeing any errors when I run the `dev` branch  - what version are you running, and are there any steps to reproduce?

jeroenkampinga:
Did some further research and found the error. I did an update from 0.13.0 to 1.1.3. Found out that in footer.html the variable `var baseurl = "{{{baseurl}}}";` was missing. You can ignore this merge request.

bmuenzenmeyer:
Glad you could figure it out - things like this is why we are taking steps to make Pattern Lab Node work as a dependency. Check this out if interested: https://github.com/pattern-lab/patternlab-node/wiki/Running-Pattern-Lab-Node-as-an-npm-Dependency

PR #270, opened 2016-02-26
Support for configuring a default pattern in config.json

geoffp:
Summary of changes:
This adds basic support for setting the default pattern to be displayed when PL is first launched. We have a need for this because we want to have a nice landing page that explains things and doesn't melt mobile devices when they try to display every pattern we have all at once.

@bmuenzenmeyer, please review and let me know if there's anything glaringly wrong here.


bmuenzenmeyer:
Oooo, interesting. I definitely like the idea and remember it coming up before. The one thing to take note of is that by altering files like `core/styleguide/js/styleguide.js` and `source/_patternlab-files/index.mustache` we are consciously deviating from potential front-end parity with PL-PHP. 

I've done it before and will probably do it again.

I think that's okay for now, FWIW, with two or three options down the road:

* Port the feature into PHP / frontend styleguide code to make it "core"
* Refactor this out as a plugin of sorts
* Fork "core" styleguide-assets-default with a node flavor

Either way, I don't see denying this PR. Will take a look later this weekend.

bmuenzenmeyer:
An addendum to the readme would be a good shortterm fix - as patternlab.io matures, this may belong there too.

geoffp:
Kind of makes the case for consolidating our UIs ASAP. How close are we right now to UI compatibility? I saw that 1.1.3 closes some PHP gaps.

bmuenzenmeyer:
I haven't dug in to check for certain, but as I see it (for some reason please convince me otherwise I am agreeable :smile:) I want to tackle the two items on https://github.com/pattern-lab/patternlab-node/wiki/Roadmap#v1xx-early-2016 first, especially finishing the file restructure. The first one is a legit gap that I want to close for feature parity too.

geoffp:
No, no -- you're right. It's probably best for the file/repo structure to be amenable to a truly pluggable UI before we do that.

Also, Pattern Engines are looking pretty good these days. We've had that branch (with a couple others mixed in, now including this one) in production for a week now, and the only hiccups we've had are with BrowserSync. Still working on that, I found a bug I need to fix and upstream. We can talk more about that later, though.

builder/patternlab.js
```diff
@@ -473,7 +473,11 @@ var patternlab_engine = function (config) {
     var viewAllPathsPartialHtml = pattern_assembler.renderPattern(viewAllPathsTemplate, {'viewallpaths': JSON.stringify(patternlab.viewAllPaths)});
 
     //render the patternlab template, with all partials
-    var patternlabSiteHtml = pattern_assembler.renderPattern(patternlabSiteTemplate, {}, {
+    debugger;
```
bmuenzenmeyer: This should be removed, yes?

bmuenzenmeyer:
@geoffp did you know [this browsersync option](https://www.browsersync.io/docs/options/#option-startPath) existed?

Me neither!

```
startPath: '/?p=pages-homepage',
```

Works like a dream.

geoffp:
Good find! I wish it did -- we need this mainly in static deployments that don't include BrowserSync.

bmuenzenmeyer:
Ah I understand - thanks for the clarification.

bmuenzenmeyer:
@geoffp poke about https://github.com/pattern-lab/patternlab-node/pull/270#discussion_r54328744

builder/patternlab.js
geoffp: Oh yes. :+1:

geoffp:
Errant debugger statement has been banished.

bmuenzenmeyer:
@geoffp this is in dev now. We will have to see how it shakes out with eventual reconciliation with [styleguide-assets-default](https://github.com/pattern-lab/styleguidekit-assets-default).

geoffp:
Cool. I'll volunteer to port it to "upstream" (we can call it that, right?) if and when the time comes.

PR #271, opened 2016-02-26
Fix double BrowserSync on whole-UI refresh after selecting pattern

geoffp:
Summary of changes:

Add glob in the BrowserSync blacklist to match ?p= pattern URLs on the main UI chrome so BS won't load on both in the case where you select a specific pattern and then refresh the whole page.

We were having mysterious issues with a double BS badge appearing unpredictably, and it turns out this was the issue.

Would you prefer I create an issue and then refer to it in cases like this, or just do the PR?

EDIT: addresses #273

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -103,7 +103,7 @@ module.exports = function(grunt) {
 					},
                     snippetOptions: {
                         // Ignore all HTML files within the templates folder
-                        blacklist: ['/index.html', '/', '/?p=*']
+                        blacklist: ['/index.html', '/', '/?*']
                     },
 					plugins: [
 						{
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -103,7 +103,7 @@ module.exports = function(grunt) {
 					},
                     snippetOptions: {
                         // Ignore all HTML files within the templates folder
-                        blacklist: ['/index.html', '/']
+                        blacklist: ['/index.html', '/', '/?p=*']
                     },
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
Nice detective work! :shipit: Excited to try this out.

Do the blacklists need to account for these query string parameters http://patternlab.io/docs/pattern-linking.html ?

Looking at that more - I wonder if we aren't setting the query string correctly since not all those values are put in the url. They seem to all respond when visiting, but they are then stripped out. Food for thought.

I don't mind the direct PR's - the only downside is losing visibility of what might be staged for next release, but I think the tweaks to CHANGELOG/ROADMAP can fix that.  I may emulate something akin to this: https://github.com/cefsharp/CefSharp/wiki/ChangeLog

geoffp:
They do not -- maybe I'll just generalize the glob to match anything starting with `/?*`. I'll have think about whether that's too risky, but I can't think of why it would be just now.

geoffp:
And, if it helps to keep track of changes, I'll create issues. It's not much trouble, and I bet it saves more time than it costs in the long run.

bmuenzenmeyer:
I'll let you test what black list values seem right, especially since you and your team can reproduce the unexpected behavior currently.

geoffp:
Oh -- I'm now realizing that I misspoke. I meant to say, "oh yeah, I had no idea those other URL patterns even existed and we definitely have to handle them."

PR #276, opened 2016-02-27
more robust parameter to json conversion

e2tha-e:
Addresses #250 (preliminary work)

Summary of changes:
This converter to well-formed JSON will correctly account for different quoting schemes, as well as quotes within quotes for parameter values. Unit tested.


bmuenzenmeyer:
Looks great - thanks for the thorough improvement and the unit tests!

PR #277, opened 2016-03-01
250 with recursion limiter

e2tha-e:
Addresses #250 

Summary of changes:
Adds recursion to the findparameters() function. However, it also limits recursion by immediately rendering non-partial Mustache tags. So if a parameter had been submitted to performed the Mustache partial inclusion, the inclusion, and therefore the recursion, would be performed, but not otherwise. This gives users the ability to code many levels of optional inclusions, but without forcing Pattern Lab to recurse every single possible option, which without the limit, would grow exponentially per level.


e2tha-e:
@bmuenzenmeyer @geoffp This is currently a Mustache/Handlbars-only solution. It escapes partial includes from rendering by temporarily switching them to ERB syntax. I think it would be trivial to extend this functionality to other templating languages, assuming we don't provide an ERB option. And even if we do, it would be trivial to escape ERB tags as well.

bmuenzenmeyer:
@e2tha-e I did a quick read through - thanks for the thorough, thoughtful improvement. I can't wait to review this!

e2tha-e:
@bmuenzenmeyer I'm looking into optimizing this further based on feedback from @vebersol in issue #250 . Feel free to review, but do not merge for the time being.

e2tha-e:
@bmuenzenmeyer Lots of merge conflicts, so not ready for final review, but here are some changes to note:

The Mustache compiler is replaced by the Hogan compiler. This reduces the memory footprint by around 50%.

processPatternRecursive no longer switches context to within each nested pattern. It stays within the context of the pattern iterated on by diveSync.

On each recursion, processPatternRecursive updates the content of a pattern with the content of the partials found therein. If there are yet more partials within the partial's content, it recurses again, until there are no more nested partials.

The pattern object has some new properties:
* .tmpTemplate stores the content of patterns when they are parsed as partials
* .escapedTemplate stores the content of patterns processed by processPatternRecursive as the template gets built out. It is called .escapedTemplate because the Mustache tags which are keyed by the pattern's JSON data are escaped from rendering by Hogan. When Hogan renders .escapedTemplate, all irrelevant Mustache tags and blocks are deleted. This prevents the possibility of infinite, and near-infinite loops, when partials are wrapped in conditional tags, and the condition is appropriately false.
* .dataKeys is an array which stores the keys in a pattern's JSON data. This array of keys is iterated through to escape tags relevant to the pattern so they are not deleted by a Hogan render.

processPatternRecursive writes to `public/patterns` at the end of each top level of recursion. Previously, processPatternRecursive would just glom onto an ever-growing patternlab object. Then, patternlab.js would write out `public/patterns` at the end of its build process, never freeing up memory until completion of the build. Now, processPatternRecursive makes sure to release no longer required memory after each pattern is processed.

I found nothing using patternlab.partials. This was both a very heavy, and redundant property. All instances have been deleted.

TO DO:
The memory footprint still gets very high on large, complicated Pattern Lab instances. An example I tested with had 268 Mustache files, 73 JSON files, and the `source/_patterns` directory totaled 1.8MB in content. With Hogan, the memory usage approached a max of 500MB during a build. It reaches this max near the end of process, quite probably when the frontend is being built. buildFrontEnd still expects a large, thoroughly glommed patternlab object for its data needs. In addition, it copies much of the patternlab object into yet more arrays for various and sundry needs. I'm sure there's a less resource-intensive way of doing this, which should be a goal for the future.

I found nothing using lineage properties built by findlineage in lineage_hunter.js. It also only works at the iterative level, and not recursively. As such, it can't do much of any value. However, if it were rebuilt to work recursively, it could be valuable for pattern-specific, targeted builds. This is really the only way to build during Gulp/Grunt watches. It's a costly waste to build every single pattern when only a few are affected.

bmuenzenmeyer:
> buildFrontEnd still expects a large, thoroughly glommed patternlab object for its data needs. In addition, it copies much of the patternlab object into yet more arrays for various and sundry needs. I'm sure there's a less resource-intensive way of doing this, which should be a goal for the future.

Completely agree on this point. 

I need to read through all this and make since of this versus the other PRs on deck.
Stay tuned.

bmuenzenmeyer:
> The Mustache compiler is replaced by the Hogan compiler. This reduces the memory footprint by around 50%.

I approve of this change!

bmuenzenmeyer:
> I found nothing using patternlab.partials. This was both a very heavy, and redundant property. All instances have been deleted.

**Good find.!**

If memory serves me correctly, lines like https://github.com/pattern-lab/patternlab-node/blob/6ec0391a427907d3bcda9b59ce16dce91483537b/core/lib/patternlab.js#L298 used to pass in the whole patternlab object, making templates like https://github.com/pattern-lab/patternlab-node/blob/6ec0391a427907d3bcda9b59ce16dce91483537b/core/templates/styleguide.mustache#L6 render

I presume taking this out, along with your other changes, still in-fact builds the frontend correctly.

bmuenzenmeyer:
> I found nothing using lineage properties built by findlineage in lineage_hunter.js

I don't quite follow what you mean here.

e2tha-e:
@bmuenzenmeyer First off, I see that there are new functions in `lineage_hunter.js`, setPatternState() and cascadePatternStates(). I haven't looked to deeply at what they do yet.

Regarding what I had observed before checking out the latest from dev, if I run `grep '\.lineage' core/lib/*`, I get a bunch of results in `lineage_hunter.js` obviously. `object_factory.js` just instantiates these properties of the pattern object:
```
core/lib/object_factory.js:  this.lineage = [];
core/lib/object_factory.js:  this.lineageIndex = [];
core/lib/object_factory.js:  this.lineageR = [];
core/lib/object_factory.js:  this.lineageRIndex = [];
```
The rest of the results are in `patternlab.js`:
```
core/lib/patternlab.js:      for (var i = 0; i < pattern.lineage.length; i++) {
core/lib/patternlab.js:        lineageArray.push(JSON.stringify(pattern.lineage[i]));
core/lib/patternlab.js:      pattern.lineage = lineageArray;
core/lib/patternlab.js:      for (var i = 0; i < pattern.lineageR.length; i++) {
core/lib/patternlab.js:        lineageRArray.push(JSON.stringify(pattern.lineageR[i]));
core/lib/patternlab.js:      pattern.lineageR = lineageRArray;
```
I see arrays getting built, but I don't see the data in them being put to use. Again, I haven't really looked too hard at setPatternState() and cascadePatternStates() yet, so I assume the data are being used there.

bmuenzenmeyer:
> I see arrays getting built, but I don't see the data in them being put to use

Lineage is used directly on the UI https://github.com/pattern-lab/patternlab-node/blob/master/core/templates/pattern-header-footer/footer-pattern.html#L6-L7

image

e2tha-e:
@bmuenzenmeyer This PR is ready for review now. It has the latest from dev merged in, and all merge conflicts resolved. Large source directories that I have from Pattern Lab PHP can be dropped in and everything just works.

geoffp:
What a phenomenal amount of work. It looks good, but there's so much that I admit I have a hard time understanding it well enough to reason about it. But I'm trying.

The main thing I'm afraid of here is that the main line of development is now on the dev-2.0 branch, which is where the pattern engines are, and it because this has lots of mustache-specific stuff in it (I think), it's gotta get into that branch somehow. @e2tha-e, would you investigate what it would take to rebase this on dev-2.0?

bmuenzenmeyer:
I echo @geoffp's admiration and concern. It'd be a big win to get this into Pattern Lab - as you've invested a lot into improving the performance of processing patterns.

If rebasing to `dev-2.0-core` is impossible, I can attempt to do so myself but it will be likely a slower slog through all these commits.

e2tha-e:
@geoffp @bmuenzenmeyer I read you loud and clear. I'll definitely make the effort to rebase into `dev-2.0-core`. I haven't looked to deeply into the `pattern-engines` or `dev-2.0-core` branches, but I'm assuming they're moving Pattern Lab Node to be template language agnostic. This is probably a good thing, because as of late, I've been looking at Pattern Lab PHP's codebase, and am thoroughly impressed with how it solves the problem of issue #250 . Drawing off of PHP's solution might actually help incorporate my solution into `dev-2.0-core`.

In the meantime, this branch has worked very well as a stopgap for the enterprise Pattern Lab work I've been doing. The real-world use-case for Pattern Lab Node over Pattern Lab PHP is for Node applications to instantiate a Pattern Lab object like any other Node object. Hopefully this can be useful for others as well.

geoffp:
@e2tha-e that's true. We are moving to support (we call it "pattern engines") multiple templating engines. Pattern parameters exist only to bolt features on to Mustache that are provided natively by other engines, so long or medium-term, I'd like to see all of pattern parameters disappear into the Mustache engine where it can't bother anyone else.

I'd love to know more about what your enterprise use case is for PL!

geoffp:
I should also warn you: dev-2.0 is pretty torn up at the moment. We're also splitting it apart into separate composable repos like the PHP version. I'll write some documentation and put it up on the READMEs.

e2tha-e:
@geoffp About the enterprise Pattern Lab implementations, I created tool called [Fepper](https://github.com/electric-eloquence/fepper) (a portmanteau of Front End Prototyper). This started out being a series of common tasks most projects would need in order to import Pattern Lab patterns into a backend, in our case, Drupal. However, more tools got added to Fepper over time, like the separation and compilation of partial JSON files for large global data sets.

Fepper originally started out using Pattern Lab PHP. The first enterprise project I used Pattern Lab PHP and Fepper for comprised 268 Mustache files. I'm currently working on a project that will likely comprise over a thousand.

So why is Pattern Lab Node better for such a large codebase than PHP? It's because since Fepper tasks are run in Node.js, it would be advantageous to instantiate Pattern Lab objects in Node.js. Fepper is built to be extended, and one of the extensions is Multisite functionality. The Multisite extension allows any number of Pattern Lab instances to be instantiated, built, and viewed within a single UI. I originally envisioned the use-case to be when building many sites that were variations on a theme. However, the actual use-case has turned out to be a way to limit expensive builds. For example, if we have 10 subsites, each with 100 patterns, for a given site, we only need to turn on the primary site and 1 subsite. This means only 200 patterns will build, instead of all 1000. It's unlikely a developer would need all subsites enabled for any given development task.

bmuenzenmeyer:
@e2tha-e Pattern Lab 2.0.0-alpha is on `dev` branch right now. This PR is crazy stale and I fear not ever merge-able.  I 100% understand a lot of this is my fault. If you find time and value in moving any of this code forward into core, [edition-node-gulp](https://github.com/pattern-lab/edition-node-gulp) and especially [patternengine-node-mustache](https://github.com/pattern-lab/patternengine-node-mustache), please let me know how I can help.

e2tha-e:
@bmuenzenmeyer I am definitely planning on working on this. I'm just trying to get the usual set of other priorities out of the way. I'm also sure you'd be interested to know that I have been asked to speak on Pattern Lab for Node.js at Node Camp in NYC this July at the UN. http://nodecamp.io/ . I accepted :-)

bmuenzenmeyer:
> I am definitely planning on working on this.

Great to hear. I hope that the new structure is more clear, but let me know if you run into any snags along the way.

> I'm also sure you'd be interested to know that I have been asked to speak on Pattern Lab for Node.js at Node Camp in NYC this July at the UN. http://nodecamp.io/ . I accepted :-)

Wow!!!!! Can't wait to here more about it.

e2tha-e:
@bmuenzenmeyer I've got a bunch of things out of the way, so I can take a look at this now. I'll have to take some time to get my bearings on the current state of things, but I'm definitely looking forward to making progress on this.

bmuenzenmeyer:
@e2tha-e great to hear. This PR's code may still be relevant, but I'd like to avoid merging in so many commits. Whatever we can do to make this several smaller, reviewable chunks I encourage it.

The codebase has shifted out from under the PR a bit too, with a lot more emphasis on modular components. I think that a bulk of code necessary will live in https://github.com/pattern-lab/patternengine-node-mustache as well as core - take some time to read through how the [core engine]*=(https://github.com/pattern-lab/patternlab-node/blob/master/core/lib/pattern_engines.js) code works with [abstractions in the Pattern Object](https://github.com/pattern-lab/patternlab-node/blob/master/core/lib/object_factory.js) and please ask me or @geoffp if you have questions.

Geoff has even started a [handlebars engine](https://github.com/geoffp/patternengine-node-handlebars) which goes a long way toward obsoleting some mustache-only syntax hurdles. Perhaps the hogan work you've done here should have its own engine too.

bmuenzenmeyer:
also - feel free to hang out in https://gitter.im/pattern-lab/node where we can all chat a bit more simply

bmuenzenmeyer:
Please see https://github.com/pattern-lab/patternlab-node/issues/603 for an explanation of why I am closing this PR.

PR #283, opened 2016-03-09
`.gitignore` changes: `Thumbs.db` and `public` directory (with all subdirectories and files).

zacanger:
Addresses #
Does not address any open issue; just addresses potential confusion over what should/shouldn't be ignored in `public`, and catches all `Thumbs.db` files.

Summary of changes:
Previously had many files/subdirectories broken out in the `.gitignore`; seems the idea is just to ignore the entire directory, which can be done with just the one line.
Not specifying subdirectory for `Thumbs.db` files will let the gitignore just catch all of 'em.


geoffp:
Thanks for this, @zacanger. @bmuenzenmeyer, I gave this a quick test, and it seems to work. Any objections?

bmuenzenmeyer:
No objections, as long as a base install works. I love the notion of cleaning stuff up like this. This ignore file is another place that showcases how ~4 years of hacking on the repo have evolved.

geoffp:
Groovy

Issue #241, opened 2016-02-03
Refactor Unit Tests


bmuenzenmeyer:
Looking through unit tests, they show the evolution of the project in all its ugly glory.
- [ ] Use addPattern() against a real file found in ./test/files
- [ ] When that's not possible, use addPattern() against a mocked pattern object
- [ ] Add helpers for unDRY components
- [ ] Reformat documents
- [ ] Standardize structure (method - condition)
- [ ] I am sure more

bmuenzenmeyer:
closing as too general

PR #284, opened 2016-03-09
Implement pattern state lowest common denominator

### LINKED ISSUE #203: Implement pattern state lowest common denominator
While reviewing documentation, I realized I never fully finished the pattern state implementation.

> Any pattern that includes a pattern partial that has a lower pattern state will inherit that state. For example, a pattern with the state of inreview that includes a pattern partial with the state of inprogress will have its state overridden and set to inprogress. It will not change to inreview until the pattern partial has a state of inreview or complete.

Code reference: https://github.com/pattern-lab/patternlab-php-core/blob/c2c4bc6a8bda2b2f9c08b197669ebc94c025e7c6/src/PatternLab/PatternData/Helpers/PatternStateHelper.php

bmuenzenmeyer:
Summary of changes:
* Pattern States are now assigned using the pattern key (atoms-colors), not the pattern name (colors). This change aligns with other uses of the key throughout Pattern Lab.
* Patterns inherit the lowest common denominator pattern state of their lineage.
* Pattern inheritance occurs only for patterns that define a pattern state themselves.

closes #203

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,8 +144,7 @@ The current selection is as follows.
 }
 ```
 ##### Pattern States
-You can set the state of a pattern by including it in `patternlab-config.json` too. The out of the box styles are in progress (orange), in review (yellow), and complete (green).
-Pattern states should be lowercase and use hyphens where spaces are present.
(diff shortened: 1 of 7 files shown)


geoffp:
Nice!

I am now realizing that I think we overlooked unit tests in our whitespace cleanup.

bmuenzenmeyer:
Right - I saw that too, in addition to eslint. It's a good bit of busy work that can be put into #241

PR #286, opened 2016-03-09
pattern engines tweaks

EvanLovely:
Here's a quick PR that should get Travis tests working again and also fixes `package.gulp.json`. Diving into some more testing too! :)


EvanLovely:
Welcome!

PR #292, opened 2016-03-14
#291: Fixes & improvements to parameter hunter

james-nash:
Addresses #291

Summary of changes:
* `core/lib/parameter_hunter.js`
   * ADDED a bunch of helper functions that `paramToJson()` now uses. Hopefully the in-source comments explain their details sufficiently well:
        * `extractNextKeyOrValue()`: Extracts the next unquoted / single-quoted / double-quoted text
        * `doubleQuoteKeyOrValIfNeeded()`: Converts text to its double-quoted equivalent
        * `extractNextKeyValPair()`: Extracts the next complete parameter (i.e. key:value pair)
    * CHANGED `paramToJson()`: Completely re-written. It now acts more parser-like, trying to find key, colon, value and comma in that order. If one of them is missing, it skips ahead until it finds a comma (or the end of the string), so that subsequent parameters can still be found.
    * CHANGED `findparameters()`: Minor change to remove the curly braces from the param string before it is passed to `paramToJson()`.
* `test/parameter_hunter_tests.js`
    * ADDED the following tests:
        * `parameter hunter parses parameters with values containing a closing parenthesis`
        * `parameter hunter parses parameters that follow a non-quoted value`
        * `parameter hunter parses parameters whose keys contain escaped quotes`
        * `parameter hunter skips malformed parameters`

I also included this unrelated change by accident (sorry):
* `.gitignore`
    * ADDED a rule to exclude `npm-debug.log`, which gets generated when you run with `debug` enabled in the config file.

Issue #291, opened 2016-03-14
Pattern parameters with non-string values cause the param values to disappear


james-nash:
I am using Pattern Lab Node `v1.2.0` on `Mac`, using the `gulp` configuration.

##### Expected Behavior
I have a mustache file which contains a pattern include with 3 parameters, the first two of which have non-string values (as they are booleans): `{{> molecules-product-search-actions( large: true, useDefaultStyle: true, prodButtonLabel: "View product" ) }}`.

I expect Pattern Lab to parse those parameters correctly and use them in the included partial.

##### Actual Behavior
When I run Pattern Lab node, the build displays warning messages along the lines of `[SyntaxError: Unexpected token u]`, but appears to complete nonetheless.

When I view the generated page, the values I have passed to the partial via the pattern parameters appear to have been lost. The included partial renders as if _all_ of its data values have been unset.

##### Steps to Reproduce
You can simulate this using the out-of-the box mustache templates that ship with Pattern Lab like so:

1. Edit `source/_patterns/02-organisms/02-comments/01-sticky-comment.mustache` and add a parameter with a non-quoted value to one of the partial includes in that file. For example: `{{> molecules-single-comment(test: 42, description: 'We are all in the gutter, ....') }}`
1. Re-build by running `grunt` (or `gulp`)
    * Note that you'll see a few `[SyntaxError: Unexpected token d]` messages on the console during the build.
1. Open the generated site in a web browser and navigated to "Pages > Article"
1. Scroll down the page to comments section and you'll see that the text and avatar image for the comment you edited have disappeared.

##### Root cause and potential fix
I did a bit of investigating and traced this bug back to the logic inside `parameter_hunter.js`. The `paramToJson()` function is incorrectly advancing the `delimitPos` by 1 for cases where the param value is not wrapped in quotes. This means the the comma after that param is removed from the remaining `paramString`, which gets processed on the next loop iteration.
Since the two regexes at the beginning of the loop expect either an opening curly brace or comma at the beginning of `paramString` they don't match. Therefore the code incorrectly assumes it has double-quoted param key and appends it to the output as is.
The net result is that `paramToJson()` returns a string with an unquoted key, which in turn generates the `[SyntaxError: Unexpected token d]` when `findparameters()` passes it into `JSON.parse()`. That then screws up the remaining logic and, well, you get the idea :-)

So, a simple fix might be to simple remove lines 105 - 107 so that `delimitPos` isn't moved on when it shouldn't be:
```js
[...]

Issue #145, opened 2015-09-10
Support HTML entities within pattern parameters


bmuenzenmeyer:
I tried putting some parentheses inside a pattern parameter string today and it did not work. We should be able to make these strings safer.  Would help with passing in things like `<strong>DANGER DANGER </strong>` too

james-nash:
I think this will get fixed by PR #293. Doing something like `{{> pattern-name( foo: '<div class="bar">Look ma, HTML code!</div>' ) }}` should then work. The only requirement would be to wrap your HTML value in quotes.
@e2tha-e : Would you agree? I could add a test for this if you like.

bmuenzenmeyer:
Great to hear!

james-nash:
Ok, just added a test to show that this issue no longer exists in PR #307

bmuenzenmeyer:
This is working like a dream - thanks you two!

bmuenzenmeyer:
We've regressed on this in v3.

JosefBredereck:
I was thinking if this is still an issue with the new feature of `{{> @partial-block }} `. 
Also if you add `{{{text}}}` to your pattern and call it with handlebars it should also work as expected `{{> pattern text="<strong>DANGER DANGER</strong>" }}`

bmuenzenmeyer:
right - this is now a won't fix! if your pattern Engine of choice doesn't support this, use a more powerful one. We won't be propping up mustache any longer

PR #293, opened 2016-03-16
250 precursor

e2tha-e:
Addresses #250 

Summary of changes:
Some edits to paramToJson in parameter_hunter.js required when parsing an enterprise Pattern Lab instance and a unit test for it.


e2tha-e:
@bmuenzenmeyer Some more precursor work for issue #250 . Please review and merge.

e2tha-e:
@bmuenzenmeyer 
cc @james-nash 
#292 seems like it's doing parallel, and possibly superior work in paramToJson. Please review, but hold off on merging, while we come to a consensus on what should be merged.

james-nash:
Makes sense. If there's anything I can do to help (e.g. if any of my code is unclear) just shout :-)

bmuenzenmeyer:
I haven't made time to review either yet - but agree this area of code is undergoing a lot of change and likely one of these two PRs needs to win out or consume the other.   I'll appreciate any degree of review you both can do in that regard in the meantime.

james-nash:
Sure. I have a bit of time this afternoon, so I'll take a closer look at the changes in this pull request.

e2tha-e:
@james-nash Could you please hold off review of commit 695471c ? I'll be submitting something more up-to-date hopefully later today. Thanks.

james-nash:
Sure thing. Just shout whenever it's ready.

e2tha-e:
@james-nash @bmuenzenmeyer Ready for review. Notes on the param parsing are in the docblock to paramToJson.

FYI, I added the JSON5 package and replaced the JSON object with it, for better error messaging on malformed JSON. Running `npm install` and `grunt nodeunit` will give a great example of its advantages.

core/lib/list_item_hunter.js
```diff
@@ -13,6 +13,7 @@
 var list_item_hunter = function () {
 
   var extend = require('util')._extend,
+    JSON = require('json5'),
```
james-nash: I wonder if it might be more appropriate to name this `JSON5`, so that people don't mistake it for the built-in `JSON` object.

core/lib/parameter_hunter.js
```diff
@@ -13,110 +13,227 @@
 var parameter_hunter = function () {
 
   var extend = require('util')._extend,
+    JSON = require('json5'),
```
james-nash: Same comment as above. I'd consider renaming to `JSON5` for clarity.

core/lib/parameter_hunter.js
```diff
@@ -13,110 +13,227 @@
...
+    //with escaped chars out of the way, crawl through paramString looking for
+    //keys and values
+    do {
 
-      //if param value is wrapped in single quotes, replace with double quotes.
-      if (paramString[0] === '\'') {
-        quotePos = paramString.search(/[^\\]'/);
+      //check if searching for a key
+      if (paramString[0] === '{' || paramString[0] === ',') {
+        paramString = paramString.substring([1], paramString.length).trim();
```
james-nash: The first `substring()` arg should be an integer, not an array (probably just a typo). It happens to work since JS coerces `[1]` to `1` in this case, but still worth tidying up to avoid confusion/bugs in future.

core/lib/parameter_hunter.js
```diff
@@ -13,110 +13,227 @@
...
+      //now, search for a value
+      if (paramString[0] === ':') {
+        paramString = paramString.substring([1], paramString.length).trim();
+
+        //since a quote of same type as its wrappers would be escaped, and we
+        //escaped those even further with their unicode, it is safe to look for
+        //wrapper pairs and conclude that their contents are values
+        switch (paramString[0]) {
+          case '"':
+            regex = /^"(.|\s)*?"/;
```
james-nash: This regex can be simplified to `/^"(.)*?"/`. The `.` matches any character, whitespace included.
Same goes for the regex below.
I made this change locally on my machine and re-ran the tests to verify that it does indeed work.

e2tha-e: @james-nash `(.)*?` doesn't work with newline characters.

core/lib/parameter_hunter.js
e2tha-e: @james-nash Great catch!

core/lib/parameter_hunter.js
james-nash: @e2tha-e Ah! OK, I missed that. Ignore that comment then.

core/lib/list_item_hunter.js
e2tha-e: @james-nash Had to debate over this myself, and I'm still torn. @bmuenzenmeyer What are your thoughts?

james-nash:
Ok, I've finished reviewing. It looks great!
There's few minor things I commented inline (and I see that you've already responded to them), but otherwise no issues.

I see that you brought across my additional test cases, so this solution successfully fixes the same problems I was attempting to fix in my PR #292. As far as I'm concerned, #292 can be rejected and this PR accepted instead.

core/lib/list_item_hunter.js
james-nash: I'm not too fussed myself. However, if you were to change it to `JSON5` then I'd suggest doing that throughout all files where you use that library.

e2tha-e:
@james-nash Thanks a million for the code review! I just pushed that typo-fix, and so am waiting on further review and merge. So hopefully issue #291 can be put to bed soon!

e2tha-e:
@bmuenzenmeyer @james-nash I'll be offline on vacation for a few days, so don't mind the radio silence too much. Looking forward to be back at it when I'm back :smile:

james-nash:
@e2tha-e: No worries. Enjoy your vacation!

core/lib/parameter_hunter.js
```diff
@@ -13,110 +13,230 @@
...
+
+    //with escaped chars out of the way, crawl through paramString looking for
+    //keys and values
     do {
 
-      //if param key is wrapped in single quotes, replace with double quotes.
-      paramString = paramString.replace(/(^\s*[\{|\,]\s*)'([^']+)'(\s*\:)/, '$1"$2"$3');
+      //check if searching for a key
+      if (paramString[0] === '{' || paramString[0] === ',') {
+        paramString = paramString.substring([1], paramString.length).trim();
```
james-nash: I found another instance of using an array instead of a number for `substring()`'s first argument. Must have missed that last time around - sorry!

core/lib/list_item_hunter.js
bmuenzenmeyer: Forgive my naivety on JSON5 (it looks like a nice solution to a problem we've all faced), but it doesn't sit right with me. My biggest worry is downstream consumption of .json5 files (I'd also argue that things like trailing commas are not good form. Yes comments would be nice). If people get accustomed to writing these files, or perhaps worse, writing them as "invalid" pure .json files, they might shoot themselves or others in the foot. Is this fear overblown? 

Another consideration in my rather scattershot comment is that it looks like they don't follow semver (yet), which *could* create problems.

... more thinking ...

Put simply, can you explain how adding JSON5 makes #250 (which PR references by name) and the parameter_hunter.paramToJson() method better? 

I am an outsider on the specifics of this issue, and have no doubt that a lot of thought has been put into getting us this far, but I need your help in understanding if this is straying too far from the goal of this issue/fix.    https://github.com/pattern-lab/patternlab-node/pull/293/commits/c8acb5ad9aebc9051ab5a2906b14b4fadcd7a55e  mentions better error handling, but in what way is it better than the existing try catch logic? http://codepen.io/bmuenzenmeyer/pen/66eb18de343ab1954abc53537fde09f0 has an example that is descriptive enough, coupled with a bit more logging on PL's part, would seem to do the trick in my mind.

Thoughts?

cc @geoffp

james-nash: It's important to note that the pattern parameter syntax is only vaguely defined here http://patternlab.io/docs/pattern-parameters.html#pattern-parameter-syntax , but since the examples have keys that are not double-quoted, it's definitely not pure JSON. It's JSON-ish :-)

I'd be in favour of keeping our parsing of the pattern parameter syntax quite lax, since it's conceivable that users might copy-paste chunks of JSON data (so then keys might be double-quoted) and, as per issue #145 it's desirable to allow HTML code to passed in as a value so allowing single- as well as double-quoted string values makes sense to me (saves you having to escape every quote around HTML attributes and stuff like that).

The changes in this PR which make use of JSON5 enable all of that, so I think they're worth keeping.

Since, I've been focussed primarily on the parsing of pattern parameters, which is in `parameter_hunter.paramToJson()`, I'm not as familiar with the other code though. Perhaps @e2tha-e can comment on that.

FWIW, where proper JSON data is expected, I'd agree that it might be better to be stricter (or at least display warnings in the build output) so as not to encourage sloppy JSON code that may not be interoperable with other tools or PatternLab versions.

e2tha-e:
@bmuenzenmeyer I fixed that typo pointed out by @james-nash . I've also renamed the instances of JSON5 to be clear that it is not the native JSON object. I'm happy to replace JSON5 entirely, but I don't think it detracts from Pattern Lab in any way. I don't think it by default encourages lax JSON in the actual JSON files. Unless you peak under the hood, how would you know? JSON5 does, however, provide much better error handling than native JSON. I initially experimented with JSON5 to see if it would it could help with converting params to JSON. I found that it wasn't necessary, but boy were those error-handling message helpful. As such, I left it in. Any thoughts or suggestions are welcome.

e2tha-e:
@bmuenzenmeyer Here's an example of the difference between native JSON and JSON5 exception handling. It's from running parameter_hunter_tests.js:

Native JSON:
```
[SyntaxError: Unexpected token ,]
```
JSON5:
```
{ [SyntaxError: Unexpected ',' at line 1 column 17 of the JSON5 data. Still to read: ",\"\":missing-key,\"\":,"]
  message: 'Unexpected \',\' at line 1 column 17 of the JSON5 data. Still to read: ",\\"\\":missing-key,\\"\\":,"',
  at: 16,
  lineNumber: 1,
  columnNumber: 17 }
```
In issue #250 , @vebersol and I mentioned we worked on enterprise Pattern Lab projects with hundreds of patterns. I know personally, that the "Unexpected token" syntax error was the bane of my existence. There's no way to lint parameters, and how was anyone expected to be able to quickly track down where that error came from?

Again, strictly syntaxed JSON is a requirement for .json files because they are read by fs.readJSONSync(). JSON5 has nothing to do with those files. I also wouldn't mind revisiting JSON5 being a drop-in replacement for JSON in the few instances where it's used, in the same manner that fs-extra is a drop-in replacement for fs.

bmuenzenmeyer:
I'm coming around to this.

I made `00-homepage.json` malformed and it silently failed, failing to apply any data to the page from the template. It's because [this](https://github.com/pattern-lab/patternlab-node/blob/c2bf904e52d3e036e11b5ab0f992b9b5ffca2c07/core/lib/pattern_assembler.js#L183). 

image

This test is probably outside the scope of the current issue/fix since it relies on `fs` - but that's the sort of error you are striving to make better. Let me give this some more thought

bmuenzenmeyer:
I'm realizing [more and more](https://github.com/pattern-lab/patternlab-node/issues/305) that users are taking PL Node into territory and scale I haven't yet experienced personally or professionally. And that there is merit to making the library more robust, even at the expense of some sense of "purity" I define in my head. Your contribution addresses a need, is thoughtfully proposed, thorough in its implementation, and covered with unit tests. What more could one ask for?

I'll need to be careful to forward port this into `dev-2.0-core`

Thanks!

PR #294, opened 2016-03-16
Remove extra '>' from body tag

tannerhodges:
A tiny fix, but there's an extra `>` in the style guide's body tag. This commit removes it.

Before:
```
<body  id="patternlab-body">>
```

After:
```
<body id="patternlab-body">
```


bmuenzenmeyer:
Thanks @tannerhodges !

PR #295, opened 2016-03-17
fix documentation links in the header bar

stevedesmond-ca:
Summary of changes:
- links in ishControls were pointing to pattern-lab.info, which is now just a Media Temple parking page
- now pointing to patternlab.io


bmuenzenmeyer:
Thanks for the fix!

PR #309, opened 2016-04-11
Fix #176 infinite reload when index.html is used and custom public/ folder

### LINKED ISSUE #176: Patternlab freaks out (loading page repeatedly) when index.html is used
I'm trying to use patternlab alongside various PHP frameworks.  I tried the PHP version, but I switched to using node because I thought it would be better integrated with the rest of the front end dev sfuff and would work for front end devs that don't have php installed on their systems, but would have node/gulp.

However, since we're running it alongside there's an index.php and an index.html -- the index.php is the entry point to the framework while the index.html is pattern lab.  We can still address it directly as index.html which initially loads the "all" page, but if I select a specific pattern the browser starts trying to load a page nonstop.

For the record, the PHP version fairs only marginal better, replacing the index.php with a ?p=<path to pattern> (which this version does too if index.html isn't used in the original address), but then on reload since it will default to index.php not index.html the ?p=<path to pattern> just loads the framework homepage with that query string (which is meaningless).

I dug through the JS a bit, but I couldn't pinpoint exactly what was going on.  Obviously the link handling is replaced and it looks like there's some custom history methods in there, but I didn't see anything which should cause it to reload like that constantly.

josketres:
Addresses #176

### DIFF
--- a/core/styleguide/js/url-handler.js
+++ b/core/styleguide/js/url-handler.js
@@ -127,7 +127,7 @@ var urlHandler = {
 	pushPattern: function (pattern, givenPath) {
 		var data         = { "pattern": pattern };
 		var fileName     = urlHandler.getFileName(pattern);
-		var expectedPath = window.location.protocol+"//"+window.location.host+window.location.pathname.replace("public/index.html","public/")+fileName;
+		var expectedPath = window.location.protocol+"//"+window.location.host+window.location.pathname.replace("/index.html","/")+fileName;
 		if (givenPath != expectedPath) {
 			// make sure to update the iframe because there was a click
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
This file goes away in PL v2.0, replaced with inclusion of the https://github.com/pattern-lab/styleguidekit-assets-default via npm.

As such I am putting this on hold until I am ready to look at it through that lens. I suspect it may be an issue via that inclusion too, but won't know for a bit yet.

bmuenzenmeyer:
[The original issues](https://github.com/pattern-lab/patternlab-node/issues/176)  no longer seems to happen using https://github.com/pattern-lab/edition-node-gulp 1.0.0 

If you are still experiencing this, please reopen and we can triage anew.

Issue #322, opened 2016-04-22
View all pages are not completely rendered


hinok:
I am using Pattern Lab Node `v1.3.0` on `Mac`, using the `gulp` configuration.

##### Expected Behavior
Generated files inside `/public/patterns` are completely rendered.

##### Actual Behavior
```
/public/patterns/00-atoms/index.html
/public/patterns/01-molecules/index.html
/public/patterns/02-organisms/index.html
```

Templates are partially rendered. You can find there a lot of mustache in index.html {{you_know_what_i_mean}}

##### Steps to Reproduce

1. Clone repository
2. Configure to use gulp
3. Run in terminal
```
gulp patternlab
```

Look that `View all` pages inside iframe have mustache, see screenshot below

![probably-bug](https://cloud.githubusercontent.com/assets/1313605/14745562/2a89567c-08ac-11e6-9456-8dd9d4ed6066.png)

bmuenzenmeyer:
Merged into `dev` thanks @hinok

PR #323, opened 2016-04-22
View all pages for atoms, molecules, organisms have still some

hinok:
Addresses #322 

Summary of changes:
Use mustache compiled header and footer sections.


bmuenzenmeyer:
A forward-port to `dev-2.0-core` may not be needed, but adding the label to make sure it's checked.

Issue #313, opened 2016-04-15
Pattern States Should Always Inherit to Consuming Patterns, Not Just When The Consuming Pattern Has a State


bmuenzenmeyer:
I am using Pattern Lab Node `v1.2.2`.

##### Expected Behavior
Lower pattern states should inherit upwards to consuming patterns, regardless of the consumer's pattern state.

##### Actual Behavior
Lower pattern states only inherit upwards to consuming patterns when the consumer has a state applied.

##### Change
Remove logic that checks for a consumer's pattern state prior to inheritance. Update unit tests.
This should be made on the 1.X path and pulled forward into 2.X

Thanks to Jesse for emailing about this issue!

bmuenzenmeyer:
This should be live in `dev` now. I will create and publish a new version soon-ish.

bmuenzenmeyer:
Forward port complete into `dev-2.0-core`

PR #331, opened 2016-05-21
Dev 2.0 core oPattern refactor

geoffp:
This PR is for code review purposes! And, to communicate that I still have one unit test assertion failing -- in the lineage hunter. I could use a hand with that; I'm not too familiar with the pattern state cascade yet.

But, this does now appear to work! There's also some light factoring out of the diveSync loop machinery so that it can be reused in the unit tests.


bmuenzenmeyer:
I think the failing pattern state cascade test is actually https://github.com/pattern-lab/patternlab-node/issues/313
Checking now

PR #333, opened 2016-05-23
Dev 2.0 Core - Engine Load

bmuenzenmeyer:
**WIP PR only** 
Track remaining work here:

- [ ] Code
- [ ] Test
- [ ] :beers:


geoffp:
Aha!

So, I've merged dev-2.0-core into this again; it's all messed up, but I think it's just because the standalone Mustache engine didn't get the Pattern-refactoring treatment. That should be easy.

Also -- load-engines does seriously break the unit testing regime. We'll have to figure something out pretty quick.

The tricky part is that now we expect engines to be installed as peers of the core, not dependencies of the core. Quick ideas:

- What if we do an edition-node-test-suite? It's maybe not ideal, but I don't hate it. And it would give us plenty of breathing room to add functional tests someday.
- We could support loading engines as dependencies as well as peers of the core; `npm test` could install them automatically, maybe? This would probably be more Travis-friendly.

bmuenzenmeyer:
> What if we do an edition-node-test-suite? It's maybe not ideal, but I don't hate it. And it would give us plenty of breathing room to add functional tests someday.

I'd  be okay with this if PRs against this repo can still reference it prior to merge. The upside of our current strategy is to encourage us _and_ the community contributions to be unit tested prior to merge. If we have to remember to run something because they are decoupled, we will inevitably forget.

> We could support loading engines as dependencies as well as peers of the core; npm test could install them automatically, maybe? This would probably be more Travis-friendly.

This sounds intriguing - but I admit I don't know much about how it all works. 

Maybe it wouldn't be so bad to use [travis' api](https://docs.travis-ci.com/user/installing-dependencies/) to install engines atop of core only during CI builds. I am pretty sure that would work - but would greatly diminish local dev testing, right? (unless of course each engine had a test suite of sorts [further duplication/fragmentation])

geoffp:
No, you're right -- making unit tests easily runnable should be a top design goal.

Oh, wow. So, okay. I just tried `npm link patternengine-node-mustache` in my working copy of core (load-engines branch) and it just works. The engine loader looks for installed engines here:
```javascript
var enginesDirectory = path.join(process.cwd(), 'node_modules');
```
So if you're running tests from core, it looks in core's `node_modules` for engines. So I guess we could just recommend that if you're hacking on core, npm install all the engines there and start running tests.

I'm pretty sure I could make it load engines flexibly from either location, so that we could include the mustache engine as a dependency of the core itself if we wanted to. If you think that's helpful!

bmuenzenmeyer:
> So if you're running tests from core, it looks in core's node_modules for engines. So I guess we could just recommend that if you're hacking on core, npm install all the engines there and start running tests.

I think this works - but it would be nice to include some documentation. I'm still fuzzy on `npm link` TBH. Do you have write access to the wiki? If so you may add any pages you wish. 


> I'm pretty sure I could make it load engines flexibly from either location, so that we could include the mustache engine as a dependency of the core itself if we wanted to. If you think that's helpful!

I like the idea of it being a dependency in core even if it feels a bit dirty or against the engine a la carte concept. My understanding of how it will manifest upon `edition-node-*` installation with npm 3 is that `patternengine-node-mustache` will be alongside core inside the edition's `node_modules` directory, accomplishing the same goal as if included at the edition level, and reducing likelihood of a user removing it optimistically only to be upset.

geoffp:
`npm link patternengine-node-mustache` here is functionally the same as `npm install patternengine-node-mustache`, but instead of downloading the package and installing in node_modules, it creates a symlink (not sure at what level that concept exists in Windows vs. Unix) in node_modules to your working copy of patternengine-node-mustache.

I actually like the robustness of being able to load engines from wherever. Doesn't even feel dirty to me.

By the way, speaking of NPM 3 -- are we doing that? And, is it safe to depend on Node 4+? We could start using some ES2105 hotness.

FWIW, I'm pretty sure that what you describe, which I think is:
- edition
 - core
 - styleguidekit-mustache-default
 - styleguidekit-assets-default
 - patternengine-node-mustache
 - patternengine-node-whatever

Is what I have set up right now, only have to do dumb stuff like require(../../node_modules/thing) instead of being able to just require(thing) peer dependencies or whatever hotness NPM3 provides that I don't understand yet.

bmuenzenmeyer:
> And, is it safe to depend on Node 4+?

I marked it 4+ yesterday at https://github.com/pattern-lab/patternlab-node/commit/b54387a0e2364ca6f4385fe64bae4d4a8da1e81d but it was mostly to shorten Travis times. I suppose that's quite lazy of me!

bmuenzenmeyer:
> Is what I have set up right now, only have to do dumb stuff like require(../../node_modules/thing) instead of being able to just require(thing) peer dependencies or whatever hotness NPM3 provides that I don't understand yet.

Ah, I see. I may have spoken out of turn or from a position of ignorance. My delight with npm v3 extended mostly from them solving the long file paths problem, especially on Windows. I wonder if [peer dependencies](https://nodejs.org/en/blog/npm/peer-dependencies/) or something else solves this `../..node_modules/` problem?

geoffp:
Just published the mustache engine and added it as a dependency of the core! Unit tests now function, though the ones written for other engines will fail. That's WIP.

bmuenzenmeyer:
Running on `edition-node-gulp#dev` I get the following

image

* only mustache loads - yay!
* error finding file that exists? - boo ?
* does render pattern trees - yay!

Does anything in https://github.com/pattern-lab/edition-node-gulp/blob/dev/gulpfile.js#L89-L97 need to change?

bmuenzenmeyer:
okay, since i was circumventing npm for my local copy of patternlab-node, i `npm link`ed core and the error no longer occurs.

image

I do however, notice that I see the pattern_engines loading output twice. is this as expected or is edition-node-gulp `gulpfile` out of date now?

bmuenzenmeyer:
@geoffp please see my question about the "double engine load" above.
I checked out travis - tests are passing minus the un-included engines. I think we can use some trickery in our [travis file](https://github.com/pattern-lab/patternlab-node/blob/dev-2.0-core/.travis.yml) to load those just in time.

geoffp:
@bmuenzenmeyer I'll check that out. I can't say that I've noticed it; I'll look for it. That's definitely not right.

PR #334, opened 2016-05-23
updating dependencies

bmuenzenmeyer:
Updating dependencies to latest thanks to `ncu`

The debatable one would be twig, since that and the other engines will be removed at some point.

@geoffp please help weigh in on that point and test if you have time.


geoffp:
No worries. The Twig dependency will indeed move to another repo, but this does no harm in the meantime!

PR #342, opened 2016-06-01
Load engines flexibly from either the current directory's node_modules

geoffp:
@bmuenzenmeyer, check this out.

This should let us load engines from the core, or from whatever the "current" directory is. This isn't a perfect solution yet, because aside from checking the core, it assumes that the you're launching grunt/gulp/nodeunit/whatever from the root of your edition and looks for engine modules there. Has anyone ever *not* done it that way? Should we be more rigorous?

### DIFF
--- a/core/lib/pattern_engines.js
+++ b/core/lib/pattern_engines.js
@@ -13,14 +13,23 @@
 var path = require('path');
 var diveSync = require('diveSync');
 var engineMatcher = /^patternengine-node-(.*)$/;
-var enginesDirectory = path.join(process.cwd(), 'node_modules');
+var enginesDirectories = [
+  {
+    displayName: 'the core',
+    path: path.resolve(__dirname, '..', '..', 'node_modules')
+  },
+  {
+    displayName: 'the edition or test directory',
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
Nice work Geoff!

bmuenzenmeyer:
> Has anyone ever not done it that way? Should we be more rigorous?

To answer your questions - no one has ever done what we are doing. Doesn't that sound dramatic! Using PL/Node core with a dependency + editions is new to the Node community, but the consumption patterns we are laying down cowpaths for now make a lot of sense, both for standalone (and perhaps) dependency consumption that we haven't imagined yet. We'll have to see I guess.

acariou:
Hi,

Here is a bit of news about our friends' wedding, just take a look here <http://gupopultu.fortheloveofthebean.com/xyncafmn>

My best to you, arnaud@neolame.com

acariou:
Hi friend! 

I've found  something really lovely and I thought  you might like it too, check it out <http://spifyriko.rubbix.com/e4ypz>

Best wishes, arnaud

PR #348, opened 2016-06-03
Hopefully fix #347. Use Path tools and regex instead of hackish

### LINKED ISSUE #347: Pattern.patternSubGroup no longer parsing correctly
I am using Pattern Lab Node `v2.0.0` on `Windows`, using the `gulp` edition.

Found while attempting to implement https://github.com/pattern-lab/edition-node-gulp/issues/27

[object_factory patternSubGroup](https://github.com/pattern-lab/patternlab-node/blob/dev-2.0-core/core/lib/object_factory.js#L40) no longer appears to report what it's comment is

For `atoms-colors` pattern, it should output `global`, but it appears to output as `Atoms\01 Global`
both in the nav

image

and in the pattern breadcrumb

image

geoffp:
Addresses #347

Summary of changes:
I'm fixing this blindly -- if this doesn't work, I'll set this up in a Windows VM and see what I can do.

Use Path tools and regex rather than the old-string length tomfoolery. Hopefully this gets us out of platform-specific path issues.

### DIFF
--- a/core/lib/object_factory.js
+++ b/core/lib/object_factory.js
@@ -37,7 +37,7 @@ var Pattern = function (relPath, data) {
   this.patternGroup = this.name.substring(this.name.indexOf('-') + 1, this.name.indexOf('-', 4) + 1 - this.name.indexOf('-') + 1);
 
   // the sub-group this pattern belongs to.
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
confirmed both instances fixed. you da best Geoff

Issue #385, opened 2016-07-10
Support foo.md inside _annotations


bmuenzenmeyer:
Much like https://github.com/pattern-lab/patternlab-node/issues/349, PL/PHP allows a user to break up multiple `.md` files inside `./_annotations/` and they will all be clobbered together.

Currently we are looking for `annotations.md` only.

bmuenzenmeyer:
Pushed to 2.2.1 on github and npm

PR #387, opened 2016-07-11
fix for issue #385 - support foo.md inside _annotations

stevemartin:
Addresses #385

Summary of changes:

* read in *.md files ( not recursive )
* clobber annotations and write
* satisfies existing tests

### DIFF
--- a/core/lib/annotation_exporter.js
+++ b/core/lib/annotation_exporter.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var path = require('path'),
+  readDir = require('readdir'),
   fs = require('fs-extra'),
   JSON5 = require('json5'),
   _ = require('lodash'),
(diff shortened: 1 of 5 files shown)


bmuenzenmeyer:
@stevemartin thank you SO much for taking the time to bake this in. It's contributions and interest like yours that keep the project going.

I noticed you added a dependency, `readdir`, which seems at first glance to offer the same functionality as an existing dependency `glob`. (This would also need be added as a dependency, not a devDependency, FWIW).

I want to accept this PR as it adds measurable improvement to the library, but I'd like to see `glob` used or `readdir` used exclusively for performance reasons. I lean toward `glob` just from our historic usage of it, but I'll leave it up to you if willing.

Thanks again!

stevemartin:
@bmuenzenmeyer doh! didn't spot `glob` before, makes total sense. I've removed the `readdir` dependency. 

Thanks for patternlab-node!

bmuenzenmeyer:
Thanks again for this @stevemartin! It will make it into next release - probably some time next week.

PR #411, opened 2016-07-28
pattern-object-schema-to-match-php

e2tha-e:
Addresses #250 precursor

Summary of changes:
In order to test large pattern sets, I needed to copy over a source dir from Pattern Lab PHP. In order use this dir without exhaustively changing tags, I needed to make these changes.
- The `.name` param required that all tildes be switched to hyphens, not just the first. I use tildes liberally in filenames. No harm switching all of them to hyphens for this sort of edge-case.
- The `patternBaseName` param required that no tildes be switched to hyphens. This is purely to match PHP's implementation.
- The `patternBaseName` param required that leading underscores in filenames be stripped. This also matches PHP's implementation.

### DIFF
--- a/core/lib/list_item_hunter.js
+++ b/core/lib/list_item_hunter.js
@@ -74,7 +74,7 @@ var list_item_hunter = function () {
           allData.link = extend({}, patternlab.data.link);
 
           //check for partials within the repeated block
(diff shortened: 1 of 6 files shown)


bmuenzenmeyer:
Please see https://github.com/pattern-lab/patternlab-node/issues/603 for an explanation of why I am closing this PR.

Issue #407, opened 2016-07-27
Maybe namechange pattern_assembler.findPartial to pattern_assembler.getPartial


e2tha-e:
The public `pattern_assembler` function is named `findPartial` while the private function is named `getPartial`. This results in a couple points of confusion. The first is trying to figure out the rationale for this difference. The second is that `findPartial` is an `engine_mustache` public function, which performs a completely different operation. I think in `pattern_assembler`, it should be named `getPartial`.

bmuenzenmeyer:
I have had the same thought a couple times - especially as the codebase has evolved and engines have a similar method.

:+1:

bmuenzenmeyer:
@e2tha-e just so you know my crazy methods, i typically leave issues open until they reach master

e2tha-e:
@bmuenzenmeyer Oops. Sorry. Will keep mind next time!

PR #413, opened 2016-07-29
Find partial to get partial

e2tha-e:
Addresses #407 

Summary of changes:
@bmuenzenmeyer 
Mostly just simple name changes. This helps when grepping for one function vs the other.

Also includes minor updates to the getPartial function itself. One avoids string-concatenation and the hard-coded "mustache" extension. Another replaces `throw` with `console.error` because `throw` throws misleading messages when the error is nothing more than a match not being found.


bmuenzenmeyer:
A thorough PR that aligns with the roadmap - thanks !

Issue #408, opened 2016-07-27
Question: are underscore-prefixed directories invalid?


e2tha-e:
@bmuenzenmeyer @dmolsen 

https://github.com/pattern-lab/patternlab-node/blob/master/core/lib/patternlab.js#L39
https://github.com/pattern-lab/patternlab-node/blob/master/core/lib/patternlab.js#L63

These two lines of code will cause Pattern Lab Node to not process files in underscore-prefixed directories. I've read through the official Pattern Lab docs and haven't found a mention of this. I've also had directories of this convention processed normally in Pattern Lab PHP, albeit in v1.x.x. Could you let me know if this condition is necessary? Because I'd like to be able work without this restriction.

bmuenzenmeyer:
This code is in here because I thought it was on spec - but now I cannot find reference to it either.

@dmolsen am I mistaken?

dmolsen:
@bmuenzenmeyer -

I would have said it was on spec but I can't find it in my code. Adding an `_` to a type or subtype doesn't clean it properly either. Sorry, without checking I would have agreed with you. Happy to go in the direction you want.

e2tha-e:
I feel it would be appropriate to hide underscore-prefixed directories in the UI, but the code as it is completely ignores them. I would like to be able to continue processing them whether in PHP or in Node.

dmolsen:
That would be similar to how underscores work now for individual patterns. Marked hidden for nav purposes but still available to include elsewhere.

PR #414, opened 2016-07-29
processing-but-hiding-underscore-prefixed-dirs

e2tha-e:
Addresses #408 

Summary of changes:
Processing, but hiding from the UI, underscore-prefixed directories, since they are processed in PHP, and there is nothing in the spec forbidding them.


dmolsen:
I've opened a spec issue for this just to formalize the request a little bit. I've voted but please double-check I summarized this correctly. I want to make sure we document a feature that should be implemented in both node and PHP.

pattern-lab/the-spec#27

bmuenzenmeyer:
I am accepting this - but will have to alter it a bit inside ui-rewrite branch

Issue #429, opened 2016-08-16
loadstarterkit failed with "TypeError: Path must be a string."


nilshartmann:
I am using Pattern Lab Node `v2.4.0` on `Windows`, with Node `v4.2.4`, using the `Gulp` Edition. (`edition-node-gulp-1.3.0`)

##### Expected Behavior
`gulp patternlab:loadstarterkit --kit=starterkit-XXX` should install the specified starter kit

##### Actual Behavior
The call to gulp fails with the error message `[TypeError: Path must be a string. Received undefined]`

##### Steps to Reproduce
Run ` gulp patternlab:loadstarterkit --kit=starterkit-mustache-materialdesign` or with any other starter kit.

bmuenzenmeyer:
fixed as part of 2.4.1

jcarpe:
I am using patternlab-node 2.4.1, but still experiencing exactly this issue.

```
npm run gulp patternlab:loadstarterkit --kit=starterkit-mustache-foundation

> edition-node-gulp@1.3.4 gulp /Projects/[project dir]
> gulp --  "patternlab:loadstarterkit"


Loading engines...
...scanning for engines in the core...
...scanning for engines in the edition or test directory...
- mustache engine: good to go
...done loading engines.

[11:51:59] Using gulpfile /Projects/[project dir]/gulpfile.js
[11:51:59] Starting 'patternlab:loadstarterkit'...
TypeError: Path must be a string. Received undefined
```

Node v6.0.0, Gulp local 4.0.0-alpha.2

bmuenzenmeyer:
i wonder if this fails when using the `npm run gulp` strategy versus the global install. more digging is required. 

of note too, is that 2.4.1 is no longer latest. perhaps upgrading could help

jcarpe:
No, I think I realized that I was screwing up the command. I'll close since it's working currently. 

Thanks!

PR #430, opened 2016-08-16
Fix call to starterkit manager

nilshartmann:
Addresses #429 

Summary of changes:
Pass `config` object from `patternlab` object to `starterkit_manager` as in `listStarterKits`


bmuenzenmeyer:
confirmed. introduced this with the postinstall script. thanks for finding, reporting, and fixing

Issue #438, opened 2016-08-17
Tests for Underscore, Handlebars, and Twig fail


geoffp:
I am using Pattern Lab Node `v2.4.2` on `Mac`, with Node `v4.4.7`, using the `Gulp` Edition.

##### Expected Behavior
Unit tests for all engines should run.

##### Actual Behavior
Tests for Underscore, Handlebars, and Twig fail due to missing outputFileSuffix blob in the included patternlab-config.json.

I'll do a PR to get the config that we provide by default in the gulp edition into this, as well as make that the default.

bmuenzenmeyer:
Travis can now load engines during test (at least what has been `npm publish`ed) 

https://travis-ci.org/pattern-lab/patternlab-node/builds/153287099

geoffp:
Excellent!

Issue #437, opened 2016-08-17
Pattern Links in "view all" ignore configured outputFilePaths


geoffp:
I am using Pattern Lab Node `v2.4.2` on `Mac`, with Node `v4.4.7`, using the `Gulp` Edition.

##### Expected Behavior
When using the following config, which supports the use of Underscore templates:
```json
"outputFileSuffixes": {
    "rendered": "-rendered",
    "rawTemplate": "",
    "markupOnly": "-markupOnly"
}
```
The pattern links in "View All" should end in "-rendered.html".

##### Actual Behavior
They end in ".html", and in this configuration, that gets you the rawTemplate version and breaks navigation due to the absence of essential scripts on the page in the iframe.

This appears to be due to widespread use of the precomputed patternLink value on Pattern objects. I propose to eliminate this field and replace it with a getPatternLink() method that takes the patternlab object as an argument so it can calculate the correct value on the fly based on the user's config.

mbulfair:
I am now seeing this only under each folder, individual view all are working. So for example

- atoms
  - color
     - colorpattern1
     - colorpattern 2
     - view all (works)
-viewall (broke)

geoffp:
@mbulfair, that's actually a different issue. Mine is that these links don't work:
![screen-shot-2016-08-17-at-2 57 39-pm](https://cloud.githubusercontent.com/assets/76232/17751107/2f71e9ca-648b-11e6-8010-4253c4366ff1.png)

bmuenzenmeyer:
@geoffp I think what you propose is a good way to address this issue going forward, versus what I've been attempting to do (monkey-patch replace when issues crop up)

bmuenzenmeyer:
@geoffp FWIW this is all working for me right now on `dev` (probably why I thought this was ready for the big time in general...) I wonder if it's a platform issue, not working on Macs? if so, perhaps the solution is easier that we thnk

![viewall](https://cloud.githubusercontent.com/assets/298435/17799524/f5310892-65a0-11e6-98f5-3900a92912b3.gif)

bmuenzenmeyer:
Or, is this only an issue when you deviate from the default `outputFileSuffixes.rendered` of `''` ? That would make more sense, says my rubber ducky.

bmuenzenmeyer:
Oh, I see what you are experiencing now after I configure an outputFileSuffix. 

I tried monkey-patching `ui_builder.js` at line 353 with

[...]

PR #442, opened 2016-08-19
Fix patternlink handling

geoffp:
Addresses #438, Addresses #437 

Summary of changes:
In here, we change the way that Pattern object are initialized. To generate a correct patternLink, which is needed for pattern links in "view all" views to work, the Pattern object now needs access to the config, because the names of output files are now configurable. Therefore, we pass the global patternlab object into the Pattern constructor.

We also update the outputFileSuffixes blobs to the latest defaults throughout, fix a bunch of engine-specific unit tests, and do a little whitespace cleanup for good measure.

### DIFF
--- a/core/lib/patternlab.js
+++ b/core/lib/patternlab.js
@@ -1,10 +1,10 @@
-/* 
- * patternlab-node - v2.4.2 - 2016 
- * 
+/*
+ * patternlab-node - v2.4.2 - 2016
+ *
  * Brian Muenzenmeyer, Geoff Pursell, and the web community.
- * Licensed under the MIT license. 
- * 
- * Many thanks to Brad Frost and Dave Olsen for inspiration, encouragement, and advice. 
+ * Licensed under the MIT license.
+ *
+ * Many thanks to Brad Frost and Dave Olsen for inspiration, encouragement, and advice.
  *
  */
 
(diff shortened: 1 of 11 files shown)


core/lib/patternlab.js
```diff
@@ -355,13 +355,13 @@ var patternlab_engine = function (config) {
 
       //write the compiled template to the public patterns directory
       var patternPage = headHTML + pattern.patternPartialCode + footerHTML;
-      fs.outputFileSync(paths.public.patterns + pattern.patternLink.replace('.html', patternlab.config.outputFileSuffixes.rendered + '.html'), patternPage);
+      fs.outputFileSync(paths.public.patterns + pattern.getPatternLink(patternlab, 'rendered'), patternPage);
```
bmuenzenmeyer: @geoffp this stuff is cool !

geoffp: Thank you! :)

bmuenzenmeyer:
I made one slight change to account for info panel template `GET`s seeming to fail.
I had other errors on my test edition but it seemed to clear up with an `npm update` of styleguide-assets-default.

geoffp:
@bmuenzenmeyer, good catch! Sorry I missed that. Makes me want to run off and start doing functional testing.

bmuenzenmeyer:
I think everything is good now! If you could give 2.4.3 a spin, that would be great

geoffp:
Doin' that now

PR #445, opened 2016-08-22
Use patternExtension for _meta head and foot

Loque-:
Addresses #

Summary of changes:

Append patternExtension from config json onto head and foot includes. My first ever PR, so hope it works! = ) Hi mum!


Loque-:
I noticed a glaring mistake in my change : ) will try again soon. Sorry for the noise!

geoffp:
No worries. Keep at it! We'll be here when you get it figured out.

Issue #446, opened 2016-08-22
Support patternengine name rather than file extension in the template tab


geoffp:
This is to support display of the pattern engine name in the template tab, should there be a positive response from Dave and Brad to https://github.com/pattern-lab/styleguidekit-assets-default/issues/43.

This is one of the issues we touched on in https://github.com/pattern-lab/the-spec/issues/26.

bmuenzenmeyer:
Released in PL Node 2.4.4

PR #447, opened 2016-08-22
Support patternengine name rather than file extension in the template tab

geoffp:
Addresses #446 

Summary of changes:
Hey, it's a one-liner. We just drop the engine name into the pattern data block on the pattern's rendered html page. The UI can be updated to use it, but otherwise will ignore it.

See also:
* https://github.com/pattern-lab/styleguidekit-assets-default/issues/43
* https://github.com/pattern-lab/styleguidekit-assets-default/pull/44

Bonus: I believe this change is harmless even if support is missing in styleguidekit-assets-default. Likewise, I believe the proposed change to styleguidekit-assets-default is harmless even if this change isn't present.

### DIFF
--- a/core/lib/patternlab.js
+++ b/core/lib/patternlab.js
@@ -339,6 +339,7 @@ var patternlab_engine = function (config) {
         patternName: pattern.patternName,
         patternPartial: pattern.patternPartial,
         patternState: pattern.patternState,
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
Tested this with a local copy of https://github.com/pattern-lab/styleguidekit-assets-default/pull/44 with and without values - works as expected

Issue #448, opened 2016-08-24
clean output html in patterns infos


gael-boyenval:
I am using Pattern Lab Node v2.4.3 on Mac, with Node v6, using the Gulp Edition.

##### Expected Behavior

Our workflow: 
1) front-end dev are building dom éléments and patterns using PL node and mustache. 
2) Back-end guys then go there and pickup the dom in the pattern lab infos panels.

it would be great to get : 

* a cleaned html output in pattern infos panels (indent, spaces, lines, tab etc)
* <s>a copy to clipboard button to any panel displaying code</s> (into pattern-lab/styleguidekit-assets-default#20)

maybe with : ``require('js-beautify').html;``

a test here : https://github.com/pattern-lab/patternlab-node/pull/449

##### Actual Behavior

none of that

bmuenzenmeyer:
@gael-boyenval merged and tweaked a bit. overall a good addition!

gael-boyenval:
@bmuenzenmeyer, yes great ! :)
We should maybe add an option to the config file for the indent size. 
Every company has his own conventions
What do you think ?

bmuenzenmeyer:
I'm opposed to adding minutiae config options like this. I understand that my change might not be what all teams are used to, but their IDE should offer formatting options and ultimately the PL code pane is about readability, which to me, is less scrolling.

That's why I said [here](https://github.com/pattern-lab/patternlab-node/pull/449#pullrequestreview-1211876) that I was hoping we could pass through any found .editorconfig to your PR's code and then truly have people define it that way. To me that's the best long-term solution and for now I'd rather not add an indent property to the configuration.

PR #449, opened 2016-08-24
Clean rendered html output

gael-boyenval:
Addresses #448 

Summary of changes:

* add js-beautify dependency
* add clean into the render function 

@bmuenzenmeyer 
that's surely not the right place to do it. the tests are failing.
just for you too see. 

maybe it would be more suited to modify ``pattern_assembler.renderPattern ? ``


bmuenzenmeyer:
@gael-boyenval I think this could likely be moved to just before the writes.

Like around https://github.com/pattern-lab/patternlab-node/blob/master/core/lib/patternlab.js#L388-L395

gael-boyenval:
@bmuenzenmeyer : sorry i didn't saw th message earlier. 
Regarding the templates, i think that the user has full hand on the output, since nothing is transformed, it's just copied. 
I assumed that the user would want to keep them as they was created.

bmuenzenmeyer:
@gael-boyenval You'll see in my additional commit (now on `dev`!) that I am beautifying templates too. 

I plan to release 2.6.0 as an alpha, so I think we can give it a go for a bit and see if there are any problems.

does that sound ok?

bmuenzenmeyer:
> sorry i didn't saw the message earlier.

and this was my first time using the new review interface, so I didn't know if it'd be the same as @mentioning you

Issue #415, opened 2016-08-01
Implement: Pattern Type Documentation


bmuenzenmeyer:
https://github.com/pattern-lab/the-spec/issues/28

bmuenzenmeyer:
I think the way is paved for this now that the viewall links are all working, but I want to bump this to a future release.

ravaan:
Hey! I want to attempt, but I am not sure what Pattern Type Documentations are, any heads up to where I can read more about it?

bmuenzenmeyer:
hey @ravaan I will get back to you - a bit busy right now, but before October I will try to fill in some details

ringods:
In my naive thinking, I can add `.md` files next to each pattern to have it documented. Is this implemented or is there still outstanding work on this?

mfranzke:
@ringods I think so as well. This is a good candidate to finally agree on that it‘s implemented, especially through the latest efforts by @JosefBredereck with MRs like #1237

PR #500, opened 2016-10-03
Fix relativeDepth check

tommcc:
Summary of changes:
- Fix syntax of depth check fallback in case of files at depth 0.

When parsing files at the 0th level (e.g., markdown associated with 1st-level folders), this regex returns `undefined`. The`|| []` was meant to guard against checking `.length` of `undefined`, but was ineffective due to this missing a pair of parentheses.

Should I attempt to add a test for this, or is it simple enough to go in as-is?

### DIFF
--- a/core/lib/pattern_assembler.js
+++ b/core/lib/pattern_assembler.js
@@ -243,6 +243,7 @@ var pattern_assembler = function () {
   }
 
   function processPatternIterative(relPath, patternlab) {
+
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
Interesting find, something that seems to have been overlooked for a long time. Sounds as if this would make patternType markdown processing easier, something I attempted but ultimately left hanging as https://github.com/pattern-lab/patternlab-node/issues/415.

I see this just getting a once over and merge - no need for a unit test

Issue #469, opened 2016-09-02
Underscore-hidden patterns must be callable by shorthand partial style


geoffp:
I am using Pattern Lab Node `v2.5.1` on `Mac`, with Node `4.4.7`, using the `Gulp` Edition.

**In Handlebars, and possibly other engines**, we are currently **off-spec** with our handling of callable patterns hidden by underscore-prefix. 
##### Expected Behavior

Given a Mustache pattern at `atoms/00-somegroup/_00-foo.mustache` I should be able to call it as `{{> atoms-foo}}`.
##### Actual Behavior

I believe one could call this underscored pattern partial today, but you'd have to call it in the nasty longform way, as in: `{{> atoms-_00-foo}}`. This is the recommended workaround until we change to the correct way.

TODO:
- [ ] Reproduce and fix if necessary in Handlebars
- [ ] Reproduce and fix if necessary in Underscore
- [ ] Reproduce and fix if necessary in Twig

geoffp:
UPDATE: this works with Mustache! The only two reports of this have been in Handlebars templates!

geoffp:
Branch is https://github.com/pattern-lab/patternlab-node/tree/fix-hidden-pattern-references

PR #509, opened 2016-10-06
Fix hidden pattern references

geoffp:
Addresses #469 

Summary of changes:
This is frustratingly simple for how much time it took. Two main changes to the non-test code:
1. Make the pattern prefix matching regex aware of leading underscores in pattern file names.
1. Factor it out and write it only once.
1. Also, pre-calculate the verbose partial name in the Pattern object. I thought I could use it, but I couldn't (handlebars doesn't tolerate slashes in partial syntax)

And lots of unit tests to prove it. I created a place for shared code in the tests.


bmuenzenmeyer:
fixed a merge conflict - going to test this manually too next free time i get - kids up now - dad time

geoffp:
Groovy. Do that dad time!

geoffp:
And, thanks for doing the merge. I picked exactly the wrong time to do stuff to unit tests.

bmuenzenmeyer:
@geoffp does using lineage to go to the hidden pattern work for you?

image

It does not for me using this quick test above

bmuenzenmeyer:
I'm okay making this a separate issue too if need be. Just curious what is happening on your end.

geoffp:
@bmuenzenmeyer Just tested -- no, it's not working for me. Separate ticket?

Issue #517, opened 2016-10-07
Convert engine_handlebars_tests.js to use tap/tape


bmuenzenmeyer:
Change per #466

PR #526, opened 2016-10-08
Convert engine_handlebars_tests.js to use tape

alexbenic:
Addresses #517

### DIFF
--- a/test/engine_handlebars_tests.js
+++ b/test/engine_handlebars_tests.js
@@ -1,10 +1,20 @@
 "use strict";
+/*eslint-disable no-shadow*/
 
+var test = require('tape');
 var path = require('path');
 var pa = require('../core/lib/pattern_assembler');
 var Pattern = require('../core/lib/object_factory').Pattern;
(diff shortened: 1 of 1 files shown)


alexbenic:
It seems to fail against node v4.6. Not sure why, will take a look today.

alexbenic:
I can't reproduce the error using:
- ```node -v``` : ```v4.6.0```
- ```npm --version ``` : ```2.15.9```
- ```nvm --version ``` : ```0.32.0```

I have tested both with and without engines installed. It works as intended.

I could refactor it to use ```tap```, but conditional testing will be less... elegant, since I haven't been able to find ```test.only``` equivalent in ```tap```.

bmuenzenmeyer:
trying to manually restart that job on travis as a first step

bmuenzenmeyer:
pulled this down local - also getting an error - i guess that's a good thing

alexbenic:
I will refactor one test to use ```tap``` and see how it goes.

bmuenzenmeyer:
@alexbenic glad to see this working - one of my todos will be to convert them all to tap so as to further remove grunt from the equation. thanks for checking this out and getting it working

PR #528, opened 2016-10-08
Convert pattern_engines_tests.js to use tap/tape

### LINKED ISSUE #512: Convert pattern_engines_tests.js to use tap/tape
Change per #466

bmuenzenmeyer:
closes #512


bmuenzenmeyer:
@geoffp since this test file is a bit different than most, I'd appreciate a second glance when you have 2 minutes

PR #533, opened 2016-10-12
Added patternlab-pattern-before-data-merge event to enable manipulati…

DavidEmanuelsen:
Summary of changes:
Added an additional event to patternlab.js to after pattern lineages have been set, and before global data has been merged with pattern data. The event is needed to be able to traverse/modify pattern data  before the merge occurs. 

We were trying to create a plugin similar to the data-inheritance plugin for php when we found there was no suitable event to hook onto. If the naming is incorrect according to any standard defined, I'll happily change it.

### DIFF
--- a/core/lib/patternlab.js
+++ b/core/lib/patternlab.js
@@ -1,10 +1,10 @@
-/* 
- * patternlab-node - v2.6.0-alpha - 2016 
- * 
+/*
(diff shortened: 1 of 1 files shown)


bmuenzenmeyer:
A welcome addition - thanks for your interest in Pattern Lab plugins!

Issue #540, opened 2016-10-21
Implement Build Algorithm for Only Changed/Affected Patterns


bmuenzenmeyer:
An idea has been floating around for a while to walk pattern lineage or reverse lineage when a pattern is changed, instead of rebuilding the whole pattern tree.

This is a large effort- but discussion can start here and move to implementation as we go.

Related concept: https://github.com/pattern-lab/patternlab-node/issues/327

tburny:
Initially patternlab would build as normal, resolving the lineage and reverse lineage of each pattern.
We can then use the (reverse?) lineage to build a dependency graph of the patterns. Each node would consist of
a (pseudo-)pattern, a timestamp at which the source file (template or JSON) was last modified and a state.
When patternlab finishes, the state of each node is set to `clean` and the graph is written to a file (e.g. JSON).
Such a graph could either be stored as a matrix or via adjacency list (i.e. given a pattern, have an array of all other patterns it depends upon) (-> TBD).

On the next run, patternlab would read the file and compare the modification dates to the moditication timestamps of each file. 
(Corner case: When a pattern is added or deleted, the graph will be updated accordingly by adding or removing edges to dependant patterns. So we need to check for differences in the file list.)
If the modification timestamp of the file is greater than the one in the graph, its node will be marked as `dirty`. 
Now we simply need to build the [transitive closure](https://en.wikipedia.org/wiki/Transitive_closure) of all dirty nodes to know which patterns we have to process.
I'll be using `A -> B` to denote that pattern `A` includes `B` as partial pattern.
So given `A -> B -> C -> D`, if `C` is `dirty` so are `B` and `A`. We do not need to recompile C and can skip it completely, because it has no dependencies.

For example you might think of it as `A` is Template, `B` is an Organism, `C` is a Molecule and D is an Atom. When rebuilding a changed molecule, no atoms are harmed.

When a pattern is compiling, we set its state to `compiling` and to `clean` when done. This way we know when something went wrong.

Still, we need to know the compilation order of all `dirty` nodes, so we apply a sprinkle of [Topological sorting](https://en.wikipedia.org/wiki/Topological_sorting) to the transitive closure to determine which templates we need to compile next.
The difference to the lineage is that here the pattern dependencies are known: 
Given `A -> B ` and `A -> C` where `B` and `C` are `dirty`, we first need to build `B` and `C` before we can rebuild `A`. 
A valid topological sorting therefore would be `C, B, A` or `B, C, A`.
 
[...]

PR #545, opened 2016-10-28
#540 only rebuild changed templates via modification date (prototype)

tburny:
Addresses #540 

Summary of changes:
This is a prototype for recompiling any templates that have been modified since they have been previously built. This works by inspecting the source and output files' modification timestamps.
A pattern is marked as `needs rebuild` if any of its source files was modified after the modification timestamp of the output file. When the template was built and the output was written to disk, it is marked as `clean` to prevent building templates more than once.

### Working
A pattern is recompiled iif

- the template itself was modified
- its data (companion json) was modified
- it is a pseudo-pattern and the according json file was modified
- existing unit tests are running

### Not working yet

- [ ] Walking transitive dependencies for recompiling
- [ ] When using grunt, `patternlab:serve` will wipe the output folder.
- [ ] missing unit tests

### Won't do (for now)
When a JSON file is changed, currently this would trigger rebuilding all transitive templates. However this is a valid use case when it comes to data inheritance.

### DIFF
--- a/core/lib/object_factory.js
+++ b/core/lib/object_factory.js
@@ -1,6 +1,8 @@
 "use strict";
 
 var patternEngines = require('./pattern_engines');
+var fs = require('fs');
+var _ = require("lodash");
 var path = require('path');
 var extend = require('util')._extend;
 
@@ -63,6 +65,11 @@ var Pattern = function (relPath, data, patternlab) {
   this.lineageRIndex = [];
   this.isPseudoPattern = false;
   this.engine = patternEngines.getEngineForPattern(this);
+  // For completeness
+  this.compileState = null;
+  // The unix timestamp when the pattern was last modified
+  this.lastModified = null;
+
 };
 
 // Pattern methods
@@ -143,6 +150,13 @@ Pattern.create = function (relPath, data, customProps, patternlab) {
   return extend(newPattern, customProps);
 };
 
+var CompileState = {
+  NEEDS_REBUILD: "needs rebuild",
+  BUILDING: "building",
+  CLEAN: "clean"
+};
+
 module.exports = {
-  Pattern: Pattern
+  Pattern: Pattern,
+  CompileState: CompileState
 };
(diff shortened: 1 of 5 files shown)


core/lib/object_factory.js
```diff
@@ -1,6 +1,8 @@
 "use strict";
 
 var patternEngines = require('./pattern_engines');
+var fs = require('fs');
```
tburny: - [ ] let these vanish, they are unused

core/lib/pattern_assembler.js
```diff
@@ -535,6 +573,9 @@ var pattern_assembler = function () {
     renderPattern: function (template, data, partials) {
       return renderPattern(template, data, partials);
     },
+    check_build_state: function (pattern, patternlab) {
```
tburny: I'm not sure about the naming convention here.

core/lib/patternlab.js
```diff
@@ -343,111 +345,119 @@ var patternlab_engine = function (config) {
...
 
+    // FIXME This simply iterates over all patterns, now we need to iterate over changed files
+    // with a defined order
+
+
+    var patternsToBuild = patternlab.patterns;
+
     //render all patterns last, so lineageR works
-    patternlab.patterns.forEach(function (pattern) {
+    patternsToBuild.forEach( pattern => renderSinglePattern(pattern, head));
```
tburny: I factored out this method, because the iteration order will be made deterministic later on when walking the graph.

test/pseudopattern_hunter_tests.js
```diff
@@ -25,7 +29,7 @@ function stubPatternlab() {
   pl.patterns = [];
   pl.partials = {};
   pl.config.patternStates = {};
-  pl.config.outputFileSuffixes = { rendered: ''}
+  pl.config.outputFileSuffixes = { rendered: ''};
```
tburny: There was a missing semicolon.

PR #546, opened 2016-10-28
#540 only rebuild changed templates via modification date (prototype)

tburny:
Addresses #540 

Summary of changes:
This is a prototype for recompiling any templates that have been modified since they have been previously built. This works by inspecting the source and output files' modification timestamps.
A pattern is marked as `needs rebuild` if any of its source files was modified after the modification timestamp of the output file. When the template was built and the output was written to disk, it is marked as `clean` to prevent building templates more than once.
### Working

A pattern is recompiled iif
- the template itself was modified
- its data (companion json) was modified
- it is a pseudo-pattern and the according json file was modified
- existing unit tests are running
### Not working yet
- [ ] Walking transitive dependencies for recompiling
- [ ] When using grunt, `patternlab:serve` will wipe the output folder.
- [ ] missing unit tests
- [] Handling of disconnected subgraphs (e.g. unused atoms etc.)

### Won't do (for now)

When a JSON file is changed, currently this would trigger rebuilding all transitive templates. However this is a valid use case when it comes to data inheritance.

### DIFF
--- a/core/lib/pattern_assembler.js
+++ b/core/lib/pattern_assembler.js
@@ -92,6 +92,7 @@ var pattern_assembler = function () {
 
   function addPattern(pattern, patternlab) {
 
+    patternlab.graph.add(pattern);
     //add the link to the global object
     patternlab.data.link[pattern.patternPartial] = '/patterns/' + pattern.patternLink;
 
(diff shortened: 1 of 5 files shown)


DavidEmanuelsen:
+1 (not tested this, but would do wonders for our productivity)

PR #547, opened 2016-10-30
Hide "All" nav item if JSON flag is set

tommcc:
Combined with pattern-lab/styleguidekit-assets-default#53 , the existing JSON flag `ishControlsHide.views-all` can be respected (hiding the "All" nav item at the topmost level).

Summary of changes:

More specifically, this allows the nav template to access the hiding options set in the JSON `ishControlsHide` object. Previously this object was only accessible to the ish template.


bmuenzenmeyer:
Pending a discussion going on at https://github.com/pattern-lab/styleguidekit-assets-default/pull/53

PR #550, opened 2016-11-02
Skip the partial expansion for listitems when using the handlebars engine

bleathem:
Summary of changes:
The `list_item_hunter.js` implementation assumes one is using the mustache engine, and fails when nesting handlebars patterns in the list tag.  This PR skips the partial expansions when the handlebars rendering engine is used, since handlebars is able to render nested partials itself.

### DIFF
--- a/core/lib/list_item_hunter.js
+++ b/core/lib/list_item_hunter.js
@@ -73,8 +73,17 @@ var list_item_hunter = function () {
           allData = plutils.mergeData(allData, itemData !== undefined ? itemData[i] : {}); //itemData could be undefined if the listblock contains no partial, just markup
           allData.link = extend({}, patternlab.data.link);
 
-          //check for partials within the repeated block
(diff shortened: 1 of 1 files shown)


bleathem:
A possibly more robust approach to solving this problem can be seen with this handlebars helper (implemented as a patternfly-node plugin):
https://github.com/bleathem/patternlab-plugin-handlebars-helper-listitems

It requires a syntax change, but delegates the list item iteration and rendering to the handlebars engine, rather than trying to hack the mustache approach as attempted in this PR.

tburny:
Maybe it would be good to have a unit test here, too? :)

bmuenzenmeyer:
@bleathem with the plugin in mind - can this PR be closed?

bleathem:
Yes, the handlebars helper approach is both less of a hack and a more complete solution than this PR.  Once https://github.com/pattern-lab/patternlab-node/pull/555 is merged and released, the plugin will be generally consumable.

Closing this PR.

PR #555, opened 2016-11-08
Expose the pattern engines as a property of the patternlab object

bleathem:
When developing a patternlab-node plugin, I found it useful to have access to the loaded pattern engines as a property of the patternlab object.  This PR introduces an `engines` property to the `patternlab` object.

### DIFF
--- a/core/lib/patternlab.js
+++ b/core/lib/patternlab.js
@@ -18,7 +18,8 @@ var diveSync = require('diveSync'),
   inherits = require('util').inherits,
   pm = require('./plugin_manager'),
   fs = require('fs-extra'),
-  plutils = require('./utilities');
+  plutils = require('./utilities'),
+  patternEngines = require('./pattern_engines');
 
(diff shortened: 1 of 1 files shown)


bleathem:
Here's an example of how I consume the patternlab.engine property in my patternlab-node plugin:
https://github.com/bleathem/patternlab-plugin-handlebars-helper-listitems/blob/master/index.js#L13

bmuenzenmeyer:
@geoffp poke

geoffp:
I have received the poke! I'll check into this.

geoffp:
Yeah, I'm cool with it. I had designed the pattern_engines module to be addressable as an array of engines when you require it, just the way I think you want, but being that it's part of patternlab-node, I can see how that would be super awkward from inside a plugin. No harm in hanging another reference on the main PL object.

bmuenzenmeyer:
This will be released with next version - thanks!

PR #558, opened 2016-11-14
Fixed the path to the plugins object. Issue 91 on the edition-node-gulp repo

giuseppepaul:
Minor change to fix this issue: https://github.com/pattern-lab/edition-node-gulp/issues/91
Updated the path to the plugins object.

### DIFF
--- a/core/lib/ui_builder.js
+++ b/core/lib/ui_builder.js
@@ -536,7 +536,7 @@ var ui_builder = function () {
     output += 'var viewAllPaths = ' + JSON.stringify(patternlab.viewAllPaths) + ';' + eol;
 
     //plugins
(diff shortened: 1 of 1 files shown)


giuseppepaul:
I've updated it and it works fine :) 

`output += 'var plugins = ' + JSON.stringify(patternlab.plugins || []) + ';' + eol;`

...but before I push it, I was thinking maybe an if statement inside the buildFrontend function before exportData gets called might be cleaner?

`if (patternlab.plugins === undefined) patternlabs.plugins = [];`

What do you think?

bmuenzenmeyer:
I am okay with either solution - but prefer mine more because it's closer to the eventual output

giuseppepaul:
No probs, I've updated :) cheers

bmuenzenmeyer:
Thanks for the help @giuseppepaul

PR #564, opened 2016-11-22
Async building: pattern *loading*

geoffp:
Here's the part where we asyncify the **loading** process. Before, we had started with just the pattern processing, which is actually the less IO-intensive of the two. This is where it gets really good if you ask me.

We have:

- Async versions of all the major parts of pattern loading.
- A thin coating of async on `processPatternRecursive()` -- just enough to let us treat it as such
- Now that all parts of pattern loading are async: major consolidation of test pattern loading across unit tests via `loadPatterns()` in `test_utils.js`. The pattern is very pleasing, and might eventually have some usefulness outside of unit tests. I have an inklinkg that It could be used to better parallelize pattern building, but at the expense of the newfound freedom to include any pattern in any other pattern.


geoffp:
Looks like we have menu-building ordering problem now -- since merging the pattern ordering work from dev, the menus don't build in order anymore. I can work on that tonight, @bmuenzenmeyer, unless you'd like to take it.

bmuenzenmeyer:
hey @geoffp I wil try to peek at this but time is tight at home - so it might be a few days.

geoffp:
No worries, I got this one.

geoffp:
This branch is obsolete, so imma close this.

PR #565, opened 2016-11-24
grab chalk and use it to make the build and engine loading process pr…

geoffp:
I wanted it to be pretty, so I did stuff to it.
screenshot from 2016-11-23 23-12-03

### DIFF
--- a/core/lib/patternlab.js
+++ b/core/lib/patternlab.js
@@ -14,13 +14,21 @@ var diveSync = require('diveSync'),
   glob = require('glob'),
   _ = require('lodash'),
   path = require('path'),
+  chalk = require('chalk'),
   cleanHtml = require('js-beautify').html,
   inherits = require('util').inherits,
   pm = require('./plugin_manager'),
   fs = require('fs-extra'),
-  plutils = require('./utilities'),
-  patternEngines = require('./pattern_engines');
+  packageInfo = require('../../package.json'),
+  plutils = require('./utilities');
 
+console.log(
+  chalk.bold('\n====[ Pattern Lab / Node'),
+  `- v${packageInfo.version}`,
+  chalk.bold(']====\n')
+);
+
+var patternEngines = require('./pattern_engines');
 var EventEmitter = require('events').EventEmitter;
 
 function buildPatternData(dataFilesPath, fsDep) {
--- a/core/lib/pattern_engines.js
+++ b/core/lib/pattern_engines.js
@@ -3,6 +3,7 @@
 
 var path = require('path');
 var diveSync = require('diveSync');
+const chalk = require('chalk');
 var engineMatcher = /^patternengine-node-(.*)$/;
 var enginesDirectories = [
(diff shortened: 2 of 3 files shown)


raphaelokon:
I switched switched CLI from `colors` to `chalk` the other day as well. In future would be good to log verbosely via a CLI flag … I use a small event-driven logger for that in the CLI …

https://github.com/pattern-lab/patternlab-node-cli/blob/master/bin/utils.js#L10-L25

tburny:
I'd like to suggest replacing all console.log statements with more robust logging - with no exceptions.
This way logs become parsable by standard command line tools like cut, sed/awk, etc. We get some timings for free, to see which step takes how long.

In summary I like the idea :)

geoffp:
@raphaelokon, that looks nice, but what's the benefit of using an event bus for logging, as opposed to just calling a simple logging function directly?

@tburny, are you after some kind of delimiter that a parser can grab onto, or timestamps for every log entry, or what? Gulp does something that, and I like that it tells you how long tasks take.

raphaelokon:
@geoffp The main benefit is that the event listener can decide what to do with the logs that were emitted. I do not have to log them in place, but can decide how to pipe them into different listeners (eg. writing log files, console.log in specific format etc). 

@tburny open to formats here.

raphaelokon:
@geoffp `console.log` is also blocking on tty terminals, like OS X due to limited buffer size (I think 1KB or so)

bmuenzenmeyer:
I've endeavored to keep dependencies low - hence the current pretty ghetto implementation. I'd like feedback from users regarding additional, perhaps cosmetic dependencies.

I also like the idea @raphaelokon proposed about a better logger. Seems like a more robust way to emit and listen to events of differening importance to the user / use case.

bmuenzenmeyer:
I've come around to this - I created #566 to finish it up.
Another issue / pr could be to replace `console.log` with something better across the board.

core/lib/patternlab.js
```diff
@@ -14,13 +14,21 @@ var diveSync = require('diveSync'),
...
   pm = require('./plugin_manager'),
   fs = require('fs-extra'),
-  plutils = require('./utilities'),
-  patternEngines = require('./pattern_engines');
+  packageInfo = require('../../package.json'),
+  plutils = require('./utilities');
 
+console.log(
+  chalk.bold('\n====[ Pattern Lab / Node'),
+  `- v${packageInfo.version}`,
```
bmuenzenmeyer: I really like what this adds to output Geoff

bmuenzenmeyer:
> @geoffp console.log is also blocking on tty terminals, like OS X due to limited buffer size (I think 1KB or so)

@raphaelokon would you recommend utilizing your event logger in this case then?

raphaelokon:
The event-based logger just delegates the logging part to the handler function(s) you call … if you call `console.log` in a handler, it would also block afaik.

Issue #376, opened 2016-06-30
EPIC: Implement Improved Pattern Metadata Parsing


bmuenzenmeyer:
Implement the following per https://github.com/pattern-lab/the-spec/issues/16
- [x] parse state out of a pattern's .md Front Matter and map to a pattern's patternState proprerty
- [x] remove `setPatternState()` from pattern_assembler
- [x] parse order out of a pattern's .md Front Matter and use to override any file order prior to rendering
- [X] parse title out of a pattern's .md Front Matter and use to override the filename 
- [x] parse additional keys ouf of a pattern's .md Front Matter
- [x] display additional keys inside {{patternDescAdditions}}   -- moved to ExtraOutput in data
- [ ] parse hidden out of a pattern's .md Front Matter and use to hide pattern from all direct individual or subtype rendering similar to _underscoredPattern.mustache today. These patterns should still be include-able as partials
- [ ] parse ~excludeFromStyleguide~ `noviewall` (or a better name) out of a pattern's .md Front Matter and use to omit pattern from all rendering
- [ ] parse `tags` out of a pattern's .md Front Matter and ... ??? (available in a future search enhancement?)
- [ ] parse `links` out of a pattern's .md Front Matter and display in a future are of the front-end
- [ ] unit test coverage

tburny:
What is the state of this?

bmuenzenmeyer:
The above list is still accurate, with my taking potshots at different items every so often. As you can see, there is some diminishing returns as you get down the list, as we haven't fully realized a usage of some metadata - but things like `order` and `state` were no-brainers because they removed metadata from the filename itself. You can say the same about `hidden`, whih is the next we should tackle.

jdagerot:
According to the discussion on gitter we could have a discussion about if the code should store all frontmatter keys no matter of the keynames and values, or really just parse a few.

I believe it would be much more flexible to just store the whole parsed markdown object into the pattern, so everyone easily can implement their own keys.

[I made a pull request about this #676](https://github.com/pattern-lab/patternlab-node/pull/676)

EvanLovely:
I've implemented several of these over in the PHP version here: https://github.com/pattern-lab/patternlab-php-core/pull/137

[...]

PR #567, opened 2016-11-25
Features/376 patternorder

bmuenzenmeyer:
Part of #376

Summary of changes:
 - Users can now specify the `order` property within a pattern's optional markdown file, which, if present, will trump the filename. 
- This lets users get away, in part with the `00-filename.mustache` hack we have had since day 1 of Pattern Lab. 
- **Note:** Directories are still beholden to their name, even if you specify an `order` property within subtype markdown. (this is a good next step, ain't it?)


geoffp:
Nice!! I love this. This will move us closer to fully operationalized patterns, since I'm sure nobody wants to call their final components 00-button.whatever.

bmuenzenmeyer:
The rest of this epic will help drive it home too!

PR #573, opened 2016-12-02
Sort the patterns array before building navigation!

geoffp:
This seems to fix menu order!

### DIFF
--- a/core/lib/pattern_assembler.js
+++ b/core/lib/pattern_assembler.js
@@ -11,6 +11,7 @@ var path = require('path'),
   lih = require('./list_item_hunter'),
   smh = require('./style_modifier_hunter'),
   ph = require('./parameter_hunter'),
+  _ = require('lodash'),
   JSON5 = require('json5');
 
 var markdown_parser = new mp();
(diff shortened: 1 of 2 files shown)


geoffp:
If this fixes things on `async-building-fix-order`, I think that branch is pretty ready for review.

bmuenzenmeyer:
I don't think this is working as most people would desire.

Sorting all patterns first messes with the viewall view. All patterns are comingled at the point of your change, regardless of their patternType (atoms, molecules, etc).

So an `order` defined on an organism of `1` is sorted beside or before orders on atoms of `1`, `2`, `3`, `4`, `5`.  Adding a molecules with order `1` does the same. Repeated builds yield results based on async race-conditions.

> I get around this in `dev` by not sorting where this PR indicates, and instead relying on the synchronous reading of the files to help ensure our structure (bad). I only sort the pattern sub types.

I need to think through this more... but to fix this, we could try:

1. keeping the sort as-is and closing this PR. and then either...

* explicitly sorting the `patternlab.patterns` as it is built (I think you mentioned that at one point)
* deferring the sort until after groupPatterns is complete (need to determine that effort)

If you can live with a bit more suspense on getting this merged, I think it needs more bake-time.
I will work on this.

geoffp:
Figures it wouldn't be that easy. 

It might be more efficient overall if we maintain the patterns array as a sorted array, as long as we have an efficient way to do that, maybe with lodash. If we move the .md file loading into the load step instead of the iterative processing step (if it isn't there already, I don't recall), then we would know the file name sort value and the .md sort value before we insert into the patterns array.

bmuenzenmeyer:
@geoffp give this a shot. I've tested locally but a second set of eyes would be great

bmuenzenmeyer:
@geoffp not sure if you ever tested this yourself, but merging for now

PR #574, opened 2016-12-08
Fix link to help wanted and up for grabs issues

scottgruber:
Addresses #

Summary of changes:
Simple update to url for up for grabs issues. Permalink change to include help wanted segment along with the up for grabs url segment.


bmuenzenmeyer:
I like hot fox better, tbh.

Issue #584, opened 2016-12-23
Incremental builds: Flag pattern for rebuild if output does not exist yet


tburny:
I am using Pattern Lab Node `v2.7.1-alpha` on `Linux`, with Node `v4.6.2`, using the `Grunt ` Edition.

##### Expected Behavior
If the `lastModified` timestamp of the output cannot be read (because it does not exist), recompile the pattern

##### Actual Behavior
Unsure.

Issue #582, opened 2016-12-23
Incremental Builds: Changing the pattern markdown should rebuilt the pattern


tburny:
I am using Pattern Lab Node `v2.7.1-alpha` on `Linux`, with Node `v4.6.2`, using the `Grunt ` Edition.

##### Expected Behavior
When modifying the markdown file of a pattern, the pattern should be marked for rebuild.

##### Actual Behavior
Nothing is done

Issue #581, opened 2016-12-23
Incremental builds: Setting cleanPublic = true after incremental build does not recompile all patterns.


tburny:
I am using Pattern Lab Node `v2.7.0-alpha` on `Linux`, with Node `v4.6.2`, using the `Grunt ` Edition.

##### Expected Behavior
When `cleanPublic = true` is set as configuration option, PL should simply rebuild all patterns.

##### Actual Behavior
It might be that the serialized graph 
- becomes out of sync with the file system, resulting in unspecified behaviour
- that there are already compiled templates, thus the according patterns are only recompiled when modified

##### Steps to Reproduce
- Build your patternlab with `cleanPublic = false`
- Set `cleanPublic = true`
- Delete some files in the pattern output directory / modify some pattern templates
- If there still is a template output or an existing graph file, only some patterns will be rebuilt or there are errors.

Issue #580, opened 2016-12-20
Renaming pattern causes patternlab to break


bramsmulders:
I am using Pattern Lab Node `v2.7.1-alpha` on `Mac`, with Node `v4.4.1`, using the `Gulp` Edition.

##### Expected Behavior
Renaming a pattern rebuilds the pattern tree without issues

##### Actual Behavior
When renaming a pattern the pattern does not match anymore in the `dependencyGraph` and thus the builder cannot find the pattern anymore. It throws an error. The only solution is to remove the patterns folder and remove the `dependencyGraph` file. Then the build succeeds again.

##### Steps to Reproduce
See above, renaming a pattern.

PR #585, opened 2016-12-23
Bugfix/incremental builds

tburny:
Addresses #580 , #581 , #582 , #584


tburny:
Maybe I can try to investigate that within the next few days.

PR #601, opened 2017-01-28
Add pattern export all

jwir3:
**Use Case**: We use pattern-lab to create a rough design, then export all of our styles and patterns to a rails web application. This PR seeks to make the export more user-friendly by: 
  1. Allowing us to export raw patterns (.mustache files) which can be interpreted by the ruby mustache template engine.
  2. Preserving directory structure of our atoms/molecules/etc... such that we don't have _too_ much directory hierarchy, but one level deeper than just a flat directory structure.
  3. Allowing us to easily specify that we want _all_ patterns to be exported, rather than having to manually specify each one (as our designs are in flux constantly right now, and there are many changes).

Summary of changes:
  - Add capability to export all patterns via a flag in `patternlab-config.json`.
  - Add capability to preserve rough directory structure via a flag in `patternlab-config.json`.
  - Add capability to export raw patterns via flag in `patternlab-config.json`.

_Note_: This is a sort of rough start to a PR. I don't expect that it's going to be approved initially, but rather it's more of a starting point. It doesn't conform to the pattern-lab spec, although, I _do_ know PHP, and would be willing to change the pattern-lab PHP code as well as the node code (node just happens to be what we use right now), but I'm not 100% sure how to go about adding this to the spec. Typically, a proof-of-concept/starting-point implementation tends to be required before modifying a specification, anyway, so this should serve as that. (This is the reason the documentation for the new flags added in `patternlab-config.json` wasn't added).

### DIFF
--- a/core/lib/pattern_exporter.js
+++ b/core/lib/pattern_exporter.js
@@ -1,6 +1,7 @@
 "use strict";
 
 var fs = require('fs-extra');
+var path = require('path');
 
 var pattern_exporter = function () {
 
@@ -19,7 +20,7 @@ var pattern_exporter = function () {
     if (exportAll) {
       for (var i = 0; i < patternlab.patterns.length; i++) {
         if (!patternlab.patterns[i].patternPartial.startsWith('-')) {
-          fs.outputFileSync(patternlab.config.patternExportDirectory + patternlab.patterns[i].patternPartial + '.html', patternlab.patterns[i].patternPartialCode);
+          exportSinglePattern(patternlab, patternlab.patterns[i]);
         }
       }
 
@@ -31,10 +32,24 @@ var pattern_exporter = function () {
       for (var j = 0; j < patternlab.patterns.length; j++) {
         if (exportPartials[i] === patternlab.patterns[j].patternPartial) {
           //write matches to the desired location
-          fs.outputFileSync(patternlab.config.patternExportDirectory + patternlab.patterns[j].patternPartial + '.html', patternlab.patterns[j].patternPartialCode);
+          exportSinglePattern(patternlab, patternlab.patterns[j]);
         }
       }
     }
+  }
+
+  function exportSinglePattern(patternlab, pattern) {
+    var preserveDirStructure = patternlab.config.patternExportPreserveDirectoryStructure;
+    var patternName = pattern.patternPartial;
+    var patternDir = patternlab.config.patternExportDirectory;
+    if (preserveDirStructure) {
+      // Extract the first part of the pattern partial as the directory in which
+      // it should go.
+      patternDir = path.join(patternDir, pattern.patternPartial.split('-')[0]);
+      patternName = pattern.patternPartial.split('-').slice(1).join('-');
+    }
+
+    fs.outputFileSync(path.join(patternDir, patternName) + '.html', pattern.patternPartialCode);
   }
 
   return {
--- a/core/lib/pattern_exporter.js
+++ b/core/lib/pattern_exporter.js
@@ -14,6 +14,17 @@ var pattern_exporter = function () {
   function exportPatterns(patternlab) {
     //read the config export options
     var exportPartials = patternlab.config.patternExportPatternPartials;
+    var exportAll = patternlab.config.patternExportAll;
+
+    if (exportAll) {
+      for (var i = 0; i < patternlab.patterns.length; i++) {
+        if (!patternlab.patterns[i].patternPartial.startsWith('-')) {
+          fs.outputFileSync(patternlab.config.patternExportDirectory + patternlab.patterns[i].patternPartial + '.html', patternlab.patterns[i].patternPartialCode);
+        }
+      }
+
+      return;
+    }
 
     //find the chosen patterns to export
     for (var i = 0; i < exportPartials.length; i++) {
--- a/core/lib/pattern_exporter.js
+++ b/core/lib/pattern_exporter.js
@@ -42,6 +42,8 @@ var pattern_exporter = function () {
     var preserveDirStructure = patternlab.config.patternExportPreserveDirectoryStructure;
     var patternName = pattern.patternPartial;
     var patternDir = patternlab.config.patternExportDirectory;
+    var patternCode = pattern.patternPartialCode;
+    var patternFileExtension = ".html";
     if (preserveDirStructure) {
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
First off - thank you SO much for your contribution. I greatly appreciate a thoughtful, thorough PR and look forward to working with y'all.

I am starting a review of this by comparing functionality against my build of PL PHP. Stay tuned.

bmuenzenmeyer:
A short while later...

It looks like export in PHP does the following, and is considered current spec:

- [ ] exports `patterns/` preserving the file structure of the patterns
- [ ] exported patterns represent their finished - rendered, compiled HTML
- [ ] exports most of the contents of `source`
image

So to me, that's the baseline we need to shoot for or address.

> - [ ] exports `patterns/` preserving the file structure of the patterns

I think your flag covers the spirit of that feature, and feedback from the community can further determine if it needs to support pattern subgroup too. My thought would be most individuals would move them around to their liking in a production scenarion anyways.

> - [ ] exported patterns represent their finished - rendered, compiled HTML

Again, your flag covers this, and then some. Users can choose their own adventure with either raw or rendered patterns.

> - [ ] exports most of the contents of `source`

I am going to make a decsion say that this is not important. Modern toolchains make the copying of css/js/images/fonts from `source/` to `public/` or `export/` trivial. PL doesn't need to do this in my opinion. Longterm, PL is going to get out of the business of moving assets around too much.

So, that being said, I am going to give this a :+1: conceptually, and it's a welcome addition to the capabilities of the library.

I will do some testing and then conduct a more formal review.

bmuenzenmeyer:
I read through the docs and noticed one more thing:

> - [ ] If you require your patterns to be exported without your global header and footer (e.g. to export a clean molecule) do the following...

This will be something to strive for. Doing this will require either some ugly regex/substring work, a different core codepath, or something more creative. I will add it to the above. I still don't see this stopping the PR, as it's a step in the right direction.

I wonder then...

> Allowing us to export raw patterns (.mustache files) which can be interpreted by the ruby mustache template engine.

Are you removing this somehow yourself?

core/lib/pattern_exporter.js
```diff
@@ -14,18 +15,50 @@ var pattern_exporter = function () {
...
+        if (!patternlab.patterns[i].patternPartial.startsWith('-')) {
+          exportSinglePattern(patternlab, patternlab.patterns[i]);
+        }
+      }
+
+      return;
+    }
 
     //find the chosen patterns to export
     for (var i = 0; i < exportPartials.length; i++) {
```
bmuenzenmeyer: I cannot think of a need to loop through a provided array if the user has chosen to export all. I suggest making these mutually exclusive. Should be simple to add this as an else block as part of the exportAll check

core/lib/pattern_exporter.js
```diff
@@ -14,18 +15,50 @@ var pattern_exporter = function () {
   function exportPatterns(patternlab) {
     //read the config export options
     var exportPartials = patternlab.config.patternExportPatternPartials;
+    var exportAll = patternlab.config.patternExportAll;
+
+    if (exportAll) {
+      for (var i = 0; i < patternlab.patterns.length; i++) {
+        if (!patternlab.patterns[i].patternPartial.startsWith('-')) {
+          exportSinglePattern(patternlab, patternlab.patterns[i]);
```
bmuenzenmeyer: I'd suggest adding a check for to make sure `patternlab.patterns[i].isPattern` === `true` before calling `exportSinglePattern()` to avoid having the meta head and meta foot from being exported.

core/lib/pattern_exporter.js
```diff
@@ -14,18 +15,50 @@ var pattern_exporter = function () {
...
+    if (preserveDirStructure) {
+      // Extract the first part of the pattern partial as the directory in which
+      // it should go.
+      patternDir = path.join(patternDir, pattern.patternPartial.split('-')[0]);
+      patternName = pattern.patternPartial.split('-').slice(1).join('-');
+    }
+
+    if (patternlab.config.patternExportRaw) {
+      patternCode = pattern.extendedTemplate;
+      patternFileExtension = "." + JSON.parse(pattern.patternData).patternExtension;
```
bmuenzenmeyer: I have no idea what that JSON.parse is doing in there.
 
`pattern.patternExtension` suffices.

jwir3:
@bmuenzenmeyer Sure thing. Thanks for the review, I'll look at it as soon as I have a chance.

bmuenzenmeyer:
@jwir3 how is it going? Any chance this work could be picked back up - or would you like me to take  the baton?

jwir3:
@bmuenzenmeyer My apologies. I got distracted and forgot about this. I should have some time this week to finish this up. I'll get it to a point we discussed. Sorry for the delay.

bmuenzenmeyer:
hello @jwir3 - still interested in finishing this? Just let me know if you would like me to pick it up

bmuenzenmeyer:
@J-Gonzalez thanks for the feedback. This PR remains important to me, but it's been waiting on response from the OP for a bit. I will likely take up the comments soon to get it into the codebase proper.

Glad its of value to you!

jwir3:
Yeah, I'm really sorry. I've been remiss in keeping this up to date. I just don't think I have the time to spend on it to really push it over the finish line. I would appreciate any help you guys could give!

bmuenzenmeyer:
@jwir3 no worries. believe me, I understand we all operate at the speed of open source (make what you will of that). 

Your initial work is already of value, and it's good to hear from you again!

core/lib/pattern_exporter.js
J-Gonzalez: There is no ```pattern.patternExtension``` - "patternExtension" only lives within ```pattern.patternData``` - which does seem lto be a stringified JSON object (hence the parse @jwir3 added).

We do have a ```pattern.fileExtension``` available (which shows up with the period in it, such as ```.hbs```. I'm guessing we could use this to clean it up, but I'm not sure to why we have this difference between fileExtension and patternExtension, and why patternExtension is only available in that weird string JSON patternData.

bmuenzenmeyer:
[1005 days later](https://www.wincalendar.com/Calendar/Date/January-27-2017)....

this is good work @jwir3 - glad it will finally see the light of day!

Issue #610, opened 2017-02-06
Make core load any pattern engines more intelligently


bmuenzenmeyer:
Part of https://github.com/pattern-lab/patternlab-node/issues/603

I am going to take this issue, as I believe it also encompasses fixing the merge conflict inside #608 

- [x] Merge #608 
- [x] Defer load of engines inside module scope until absolutely necessary within build
- [x] Fix unit tests
- [x] Pass or alter linting
- [x] Confirm watches still work

bmuenzenmeyer:
Descoped #617 from this effort, citing timing.

PR #608, opened 2017-02-05
Change var to const or let

benhalverson:
Addresses #598 

Summary of changes:
This is the start of converting var's to const or let. 
Looking for some initial feedback

PR #616, opened 2017-02-08
Deferred engine load

bmuenzenmeyer:
Addresses #610

Summary of changes:
 - Allows the caller to determine when to load all engines, removing this check from the module scope of `pattern_engine.js`. This should help control the experience working with core a bit more.
- This also includes a merge for [#608 Change var to const or let](https://github.com/pattern-lab/patternlab-node/pull/608) which encompassed a bunch of stellar work by @benhalverson  

There is still one broken unit test, but I intend to look into that separately.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -83,6 +83,10 @@
     "space-infix-ops": 1,
     "valid-typeof": 2,
     "vars-on-top": 0,
-    "wrap-iife": [2, "inside"]
+    "wrap-iife": [2, "inside"],
+    "prefer-const": ["error", {
+      "destructuring": "any",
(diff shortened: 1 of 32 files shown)


bmuenzenmeyer:
One thing I am not sure of is why I don't see the commits from @benhalverson from #608. I had to merge that manually which led to this work, but I had expected them to show up here. Any ideas why they aren't reflected? I want you to get credit for those commits. Does this bother you at all Ben?

bmuenzenmeyer:
@raphaelokon can you retry your CLI against `dev-3.0` ?

raphaelokon:
@bmuenzenmeyer Sure will do. Thanks.

PR #626, opened 2017-02-21
Option of excluding sub pattern types from the styleguide

AndrewLeedham:
Addresses: Option of excluding sub pattern types from the styleguide

Summary of changes:
* Replicated omitPatternType for 'patternType/patternSubtype'
* Duplicated typePatterns as styleguideTypePatterns to seperate stylguide and nav menu creation.

### DIFF
--- a/core/lib/ui_builder.js
+++ b/core/lib/ui_builder.js
@@ -515,7 +515,7 @@ var ui_builder = function () {
           });
         if (omitPatternType) {
           if (patternlab.config.debug) {
-            console.log('Omitting ' + patternType + '/' + patternSubtype + ' from  building a viewall page because its patternGroup is specified in styleguideExcludes.');
+            console.log('Omitting ' + patternType + '/' + patternSubtype + ' from  building a viewall page because its patternSubGroup is specified in styleguideExcludes.');
           }
         } else {
           styleguideTypePatterns = styleguideTypePatterns.concat(subtypePatterns);
--- a/core/lib/ui_builder.js
+++ b/core/lib/ui_builder.js
@@ -484,7 +484,7 @@ var ui_builder = function () {
     _.forEach(styleguidePatterns.patternGroups, function (patternTypeObj, patternType) {
 
       var p;
-      var typePatterns = [];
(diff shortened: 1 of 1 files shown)


core/lib/ui_builder.js
```diff
@@ -508,6 +508,19 @@ var ui_builder = function () {
...
         });
 
+        //determine if we should omit this subpatterntype completely from the viewall page
+        var omitPatternType = styleGuideExcludes && styleGuideExcludes.length
+          && _.some(styleGuideExcludes, function (exclude) {
+            return exclude === patternType + '/' + patternSubtype;
+          });
+        if (omitPatternType) {
+          if (patternlab.config.debug) {
+            console.log('Omitting ' + patternType + '/' + patternSubtype + ' from  building a viewall page because its patternGroup is specified in styleguideExcludes.');
```
bmuenzenmeyer: Should this be pattern subGroup?

AndrewLeedham: You're not wrong, must of missed that, I'll commit the change now

AndrewLeedham:
@bmuenzenmeyer Thanks for the feedback and quick reply. With regards to usage examples, what do you mean by a snippet? The intention is that sub-types can be specified within the already existing 'styleGuideExcludes' array with the syntax "patternType/subPatternType". Since sub-types are user defined, adding this to the default `patternlab-config.json` wouldn't make all too much sense.

bmuenzenmeyer:
By snippet I meant how would one exclude a subtype.

```
...
"styleguideExcludes" : ["atoms/global"]
...
```

would do the trick if I understand correctly

AndrewLeedham:
@bmuenzenmeyer Sorry I worded that pretty badly. I meant where do you want me to include the snippet? in the default patternlab-config.json?

bmuenzenmeyer:
just here, so I can make sure to mention it in the release notes or wiki 
if i have that syntax as you'd expect, we are all good 🚀

AndrewLeedham:
@bmuenzenmeyer Ah okay fair enough. Yeah you have it right.

Issue #628, opened 2017-02-23
StarterKits should be able to override settings in patternlab-config.json


tburny:
I am using Pattern Lab Node `v2.7.2` on `Linux`, with Node `v4.6.2`, using the `Gulp or Grunt ` Edition.

By discussion with @bmuenzenmeyer:
I created a [Demo StarterKit which is using the Thymol engine](https://github.com/headissue/starterkit-thymol-demo). However the Gulp/Grunt Edition come with `mustache` as preconfigured `patternExtension`.

StarterKits should be able to overwrite fields in the Edition's `[patternlab-config.json](url)` so I can change it to `"patternExtension": "html"` automatically upon `npm install`.

When a StarterKit is loaded by Pattern Lab, it should look for a `starterkit-config.json` file, which contains all properties that have to be overridden in the edition's.

We agreed that instead of copying a whole configuration file it would be better to read the `starterkit-config.json` into `patternlab.config` and serialize the config file back to disk again - similar to how it is already done for plugins.

PR #632, opened 2017-02-24
Feature/628 starterkit config

tburny:
**Addresses #628  StarterKits should be able to override settings in `patternlab-config.json`**

Summary of changes:
1. read `starterkit-config.json`
2. Deep merge into `patternlab-config.json`. Existing object keys are overwritten, including array values.
3. Save result to `patternlab-config.json`

### DIFF
--- a/core/lib/patternlab.js
+++ b/core/lib/patternlab.js
@@ -270,7 +270,8 @@ var patternlab_engine = function (config) {
   }
 
   function loadStarterKit(starterkitName, clean) {
-    var starterkit_manager = new sm(patternlab.config);
+    const configPath = path.resolve(process.cwd(), 'patternlab-config.json');
+    const starterkit_manager = new sm(patternlab.config, configPath);
     starterkit_manager.load_starterkit(starterkitName, clean);
   }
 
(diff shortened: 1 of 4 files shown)


bmuenzenmeyer:
I fear I may have misdirected your work on this one @tburny 

You can see here how this value is used on the frontend: https://github.com/pattern-lab/styleguidekit-assets-default/search?utf8=%E2%9C%93&q=patternExtension

To me, it makes a lot more sense to have the patternengine potentially set this value, if anything. In a user's workflow, it feels like the engine provided more utility over the starterkit in dictating a change to this value, if anything (it's likely a one-time edit).

Are there other keys that you can envision a starterkit needing to overwrite/merge upon installation?

tburny:
> Are there other keys that you can envision a starterkit needing to overwrite/merge upon installation?
Just to be clean what the vision is: As a user, I first choose a plain PL or Gulp/Grunt edition. Then I'll choose a starter kit and load it.
Each starter kit can come in different flavours:

- Different template engine (template extension)
- Different source paths (e.g. `javascript` instead of `js`, `img` instead of images, etc.)
- Different pattern group naming ( e.g. `dna`, `cell`, `organism`) resulting in other names for `styleguideExcludes`
- Using LESS instead of SASS: `ignored-directories: ["less"]`

I think the StarterKit should have the flexibility to meet a user's desires :) Also, if you have 3 pattern engines installed, with mustache being already installed as part of PL, how would you determine the `patternExtension` required for a certain StarterKit?

Issue #643, opened 2017-03-22
Documentation generation


benhalverson:
We should have a way to see the output of the jsDoc comments

bmuenzenmeyer:
closing due to https://github.com/pattern-lab/patternlab-node/pull/644

PR #644, opened 2017-03-22
Automagically generate docs

benhalverson:
Addresses # https://github.com/pattern-lab/patternlab-node/issues/643

- Summary of changes:
- Added jsDoc dev dependency
- Updated .gitignore 
- Added new command in package.json


bmuenzenmeyer:
Sorry for lack of communication on this - I want to think through how useful this is. If we expect users to peruse the source code in a fashion like this.

The output is HTML of each file, yes?

benhalverson:
@bmuenzenmeyer yes the output is html files 
you can see all the files in the /out folder which can of course be renamed to docs or something else

bmuenzenmeyer:
While I understand the impetus for this work and appreciate you taking the time to put this out there - I feel like the primary use case for looking over jsDocs is within an editor already. This isn't lodash, or some libary with a super duper public api. most of the jsdocs are internal facing in relevance. editors like vscode display the jsdocs by default - which is a huge win. other editors can likely do the same

i hope this doesnt discourage you, because I value the time you spent on this and other parts of the project. should others express an extreme desire for this functionality, we may revisit it

PR #646, opened 2017-04-04
Config access for pattern engines

geoffp:
This is the simplest and least disruptive way I can think of to make engines configurable. If supported by each engine, call a function that injects the global pattern lab config into the closure scope of the engine so that engines can make configurable decisions.

@bramsmulders and I are thinking this is necessary to be able to configure the React engine -- the first use case being the ability to shut off client-side code generation if you don't need interactive components and / or are just using React components to make a static style guide. There should be some build speed benefit to doing that. 

This might also benefit the Handlebars engine, I seem to remember that it has some things you can configure.

See https://github.com/pattern-lab/patternengine-node-mustache/pull/8 for the Mustache implementation.

### DIFF
--- a/core/lib/pattern_engines.js
+++ b/core/lib/pattern_engines.js
@@ -62,7 +62,7 @@ function findEngineModulesInDirectory(dir) {
 
 const PatternEngines = Object.create({
 
-  loadAllEngines: function () {
+  loadAllEngines: function (patternLabConfig) {
     var self = this;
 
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
I love this, the reference PR https://github.com/pattern-lab/patternengine-node-mustache/pull/8 (which I might close rather than accept) and greatly appreciate the work.

bramsmulders:
Whoop!

Issue #647, opened 2017-04-12
pattern links resolve to absolute paths instead of relative paths


bmuenzenmeyer:
I am using Pattern Lab Node `v2.9.1` on `Windows`, with Node `v6.9.5`, using the `Gulp` Edition.

##### Expected Behavior
pattern links `{{link.pages-foo}}` should resolve to a relative URL

**Example**: 
In PL PHP, for example: the [demo login](http://demo.patternlab.io/?p=pages-login) form action link [code found here](https://github.com/pattern-lab/starterkit-mustache-demo/blob/master/dist/_patterns/02-organisms/forms/stacked-form.mustache) should resolve 

```
<form class="c-form c-form--stacked" method="post" action="{{ link.pages-dashboard }}">
```

to 

```
<form class="c-form c-form--stacked" method="post" action="../../patterns/04-pages-02-dashboard/04-pages-02-dashboard.html">
```

##### Actual Behavior

In PL Node, however, this resolves to

```
<form class="c-form c-form--stacked" method="post" action="/patterns/04-pages-02-dashboard\04-pages-02-dashboard.rendered.html">
```

Leads to a broken page.

##### Steps to Reproduce

1.  Download and use the https://github.com/pattern-lab/starterkit-mustache-demo starterkit 
2. Inspect/test the login form

bradfrost:
Thanks @bmuenzenmeyer. This would be a super great thing to resolve as it provides a lot of value to teams doing a lot of user testing and prototype work. We've done this kind of thing of several projects and it works quite well.

bmuenzenmeyer:
I agree @bradfrost.  I need to dive into this closer. 

Would you agree that the demo login is the perfect test case? I will work from that if you think so. I did review @pcvonz's PR but something seemed amiss. 

I will prioritize this, thanks for your feedback!

bradfrost:
Yeah I think that's right. Either that or a basic search form that clicks through to another page on submit. They're more or less the same!

bmuenzenmeyer:
@bradfrost is this an acceptable workaround for you?

inside https://github.com/pattern-lab/starterkit-mustache-demo/blob/master/dist/_patterns/02-organisms/forms/stacked-form.mustache

```diff
- <form class="c-form c-form--stacked" method="post" action="{{ link.pages-dashboard }}">
+ <form class="c-form c-form--stacked" method="get" action="{{ link.pages-dashboard }}">
```


changing the `post` to a `get` allows the form to submit

[...]

PR #649, opened 2017-04-14
make pattern links relative

pcvonz:
Addresses #647

Summary of changes:
Just changing a line to make links relative so serving the static pages via a web server wont 404.

### DIFF
--- a/core/lib/pattern_assembler.js
+++ b/core/lib/pattern_assembler.js
@@ -96,7 +96,7 @@ var pattern_assembler = function () {
   function addPattern(pattern, patternlab) {
 
     //add the link to the global object
-    patternlab.data.link[pattern.patternPartial] = '/patterns/' + pattern.patternLink;
+    patternlab.data.link[pattern.patternPartial] = '../' + pattern.patternLink;
 
     //only push to array if the array doesn't contain this pattern
     var isNew = true;


bmuenzenmeyer:
In my initial testing, this seems to resolve normal links fine, but the form post use case in #647 is still not working. I want to look at this today if possible.

pcvonz:
It works with a form action in my own spin of patternlab-node.
I'm assuming that to get to another pattern the link needs to point one directory up relative to the current pattern.

bmuenzenmeyer:
> I'm assuming that to get to another pattern the link needs to point one directory up relative to the current pattern.

the links are not supposed to rely on user's knowlege of where one pattern is in relation to another.

specifying `{{link.pages-foo}}` should resolve to the right location, no matter where included (data, markup, flat pattern, deep pattern, etc)

That is still doesn't work is my chief concern. i need to dig in more to understand the differences between PL PHP and PL Node's handling of this exact use case. One difference is PL Node uses outputSuffixes, but those have always otherwise resolves pattern links too.

bmuenzenmeyer:
@pcvonz can you provide any more detail on the outcome of this, from the angle of the hosted output?

pcvonz:
So from what I understand: the problem is that Patternlab assumes that you're going to serve the generated directory from the root url instead of putting it behind some path. Example: `https://example.com/custom_path/`. 
Every link to another pattern is going to resolve to `https://example.com/patterns/some_pattern` instead of `https://example.com/custom_path/patterns/some_pattern`
From what my patternlab project generated, I assumed that patterns are generated into a two level heirarchy, like this:
```
|__patterns                                                                                                        
| |__00-atoms-text-inline                                            
| | |__00-atoms-text-inline.rendered.html                               
| | |__00-atoms-text-inline.mustache                            
| | |__00-atoms-text-inline.markup-only.html                  
| |__00-atoms-text-dropdown                                   
| | |__00-atoms-text-dropdown.markup-only.html                
| | |__00-atoms-text-dropdown.mustache                        
| | |__00-atoms-text-dropdown.rendered.html 
```
If that's correct, then the change I made should always resolve to the correct pattern. 

I've noticed that projects like Vue usually have an option somewhere for a custom basename. A similar option could be added to the patternlab config perhaps?

bmuenzenmeyer:
Dang you stale bot - you were supposed to warn first!

bmuenzenmeyer:
@pcvonz  OK - back at this.

I tested output by building pattern lab and then using https://www.npmjs.com/package/ecstatic#opts-basedir to simulate hosting

```
cd to/patternlab-edition
gulp patternlab:build
cd public
ecstatic ./ --port 8080 --baseDir patterns
```
this correctly serve PL from http://localhost:8080/patterns/ but links seem to be resolving wrong

`http://localhost:8080/patterns/styleguide/html/link.organisms-x`

can you verify I am testing adequately?

pcvonz:
Hmm, I initially had that problem, but then realized I was on the master branch and not dev. 

I pushed what was generated from the starter-kit-mustache demo to a gh-pages branch if you want to take a look: https://pcvonz.github.io/patternlab-node/. I think hosting from here mimics pretty well what this pull request aimed to resolve. Some of the images aren't linked correctly, but I think that's a separate issue.

That aside, I think it'd be an even better idea to add an option to the patternlab config. Then we could replace the line in this pull request to something along the lines of: 

`patternlab.data.link[pattern.patternPartial] = patternlab.config.baseDir + pattern.patternLink;`

Thoughts?

bmuenzenmeyer:
Thanks for the additional analysis. Something is fishy with a lot of assets - you are right. I would be surprised if it didn't have something to do with the subdirectory. I've seen this before.

Having a baseDir is exactly what we had pre- PL Node 2.X, but with the advent of the shared UI, it was removed as off-spec. I now think that proposing that feature and resurrecting that code both in this repo and `styleguidekit-assets-default` is the right path to go.

acarnwath:
I do not know how this works under the hood, so let me know if it is not possible, but for our situation, a base directory specified in a config file (used, I assume upon build?) will not work - our local environments our simply localhost:3000/patterns... but we then copy/paste the contents of our local public folder to a hosted server where the path is quite a bit deeper, so we need _relative_ paths for it to work in both situations.

hrkfdn:
This is still a problem, right?

Issue #611, opened 2017-02-06
Alter each engine to spawn its head and foot pattern into _meta/ if not found


bmuenzenmeyer:
part of https://github.com/pattern-lab/patternlab-node/issues/603

- [x] mustache
- [x] handlebars
- [x] twig
- [x] underscore
- [ ] ~react~ https://github.com/pattern-lab/patternengine-node-react/issues

bmuenzenmeyer:
would be great if anyone else could help me understand what Underscore or React expect for these patterns

bmuenzenmeyer:
paging @geoffp - we discussed this what feels like ages ago, but can you ping me where I can find the underscore meta patterns?

geoffp:
@bmuenzenmeyer actually, I don't think we ever made any. It looks like in our underscore project, we're still using the default-ish mustache meta patterns, even though underscore is the main engine in use.

bmuenzenmeyer:
Calling this done. future react work will need to account for this.

PR #657, opened 2017-05-03
Allow _meta patterns to support any engine

AndrewLeedham:
Addresses: `_meta` patterns being forced to use mustache.

Summary of changes:

- Replaced `".mustache"` literal with `patterlab.config.patternExtension`.
- Allows the use of other engines like handlebars for `_meta` files.


bramsmulders:
Great idea, but this wouldn't work in some engines. For instance the react engine cannot spit out an incomplete dom tree.

cloudsociety:
Curious if this _meta section would be a better fit in pattern engines themselves. This should allow engines to have better control over their respective file formats. A single method in the engine that either calls multiple meta files or one (react) wrapper file.

Possible PL3.0 wish?

bmuenzenmeyer:
@cloudsociety 
you are right on target. 🎯 

this is basically the idea behind https://github.com/pattern-lab/patternlab-node/issues/611 which is on target for 3.0

I refrained from commenting on this PR until I could think through how (or if at all) `patterlab.config.patternExtension` factors into that solution

AndrewLeedham:
This was more of a hot fix for a project I'm working on, that I thought I'd create a PR for, so I didn't consider all outcomes. I also noticed that the styleguidekit extensions are fixed as `.mustache`, so I can't use [styleguidekit-handlebars-default](https://www.npmjs.com/package/styleguidekit-handlebars-default).

cloudsociety:
@bmuenzenmeyer Awesome! Looking forward to it!

@AndrewLeedham I'll still steal this bit of code so my patterns can build under a single engine in 2.x (having too much fun with Twig/Nunjucks to return to Mustache). Thanks for the share.

-Dave

bmuenzenmeyer:
since this is in the spirit of what we want to do, but the wrong implementation, i am going to close this for now. PL Node 3 will handle this with my above approach

PR #668, opened 2017-07-05
Fix parameter inheritance

tklae:
**Context:**
We are using patternlab together with [plugin-node-data-inheritance](https://github.com/Altinn/plugin-node-data-inheritance), so the following problem might not occur on "pure" patternlab installations:
1. Define a simple pattern `<p>{{foo}} {{bar}}</p>` and define some json data for it (let's say properties `{'foo':'default', 'bar':'value'}`). The pattern will be rendered using the data provided: `<p>default value</p>`
2. Define another pattern that includes the first pattern.
3. Pass only one parameter from the second pattern into the included pattern `{{> p-inner(foo: "overridden")}}`. The pattern will be rendered using the parameter passed in, but every missing property will be empty unless it's defined on the outer pattern. Rendered result: `<p>overridden </p>`

This behavior leads to us always having to define the full data set either as parameters or in the including pattern's properties and we cannot override just individual values.

**Summary of changes:**
- Extended parameter_hunter to merge properties in the following order: global properties -> included partial data (wasn't previously considered) -> includer partial data -> parameter data
- Added tests for 3 different scenarios, including data being defined on all levels. While the last test is probably enough to test the full behavior, it's also a bit harder to understand than the individual tests.


dennisfrank:
@bmuenzenmeyer Is is there any chance that this PR will be merged?

bmuenzenmeyer:
@dennisfrank I just tested this use case against `master` and it worked as expected

image


[source.zip](https://github.com/pattern-lab/patternlab-node/files/1279978/source.zip)

please confirm I am recreating the use case correctly

if true, I suspect the plugin is interferring with intended behavior?

PR #674, opened 2017-08-03
Fixed Slack channel link

iansoper:
Addresses #

Summary of changes:


bmuenzenmeyer:
btw... this was fixed in https://github.com/pattern-lab/patternlab-node/pull/666 but not yet released, entirely my fault.  i imagine this will cause a merge conflict when `dev` reaches `master` next.

I should be a better steward of these changes - I know. 

My move and job change are settling down, and there is certainly plenty to do!

Issue #559, opened 2016-11-14
Pattern subdir property contains incorrect directory separator on windows


sandrojohanides:
I am using Pattern Lab Node `v1.3.4` on `Windows`, with Node `v6.9.1`, using the `Gulp` Edition.

##### Expected Behavior
Pattern object subdir property should contain forward slash `/`

##### Actual Behavior
Pattern object subdir property contains backslash `\\` showing an error `Could not find pattern referenced with partial syntax ' + partialName + '. This can occur when a pattern was renamed, moved, or no longer exists but it still called within a different template somewhere.`

##### Steps to Reproduce
Try to include partial in mustache template using Default Include Syntax `{{> dir/subdir/partial }}` on windows

somatonic:
Hi, I´m working with PL on MacOSX and now tried my project on Windows and run into this bug. Ißm also using verbose mustache include syntax. 

I investigated and found it´s a dir separator issue. I added a line in pattern_assembler.js to get it working, I´m sure this is not the correct fix but gives a hint.

´´´
    //else look by verbose syntax
    for (var i = 0; i < patternlab.patterns.length; i++) {
      switch (partialName) {
        case patternlab.patterns[i].relPath:
        case patternlab.patterns[i].subdir.replace("\\","/") + '/' + patternlab.patterns[i].fileName:
        case patternlab.patterns[i].subdir + '/' + patternlab.patterns[i].fileName:
          return patternlab.patterns[i];
      }
    }
´´´

Now when I get that fixed, there´s a new issue with those partial includes. It  only allows me to include a pattern that is further up in the file ordering. So a structure like

00-pattern
  pattern1
  pattern2


I can´t include pattern2 in pattern1. Once I change it to different order it works like:

00-pattern
   _pattern2
   pattern1

This happened to me only on Windows using verbose include syntax once I added my fix. It seems that when using the shorthand syntax it works fine. Not sure what is happening. Thanks, any help appereciated to fix this.

sandrojohanides:
@somatonic we (seemingly) managed to fix it with adding 2 lines in `\node_modules\patternengine-node-mustache\lib\engine_mustache.js"`:

`
 22 var path = require('path');
113 foundPatternPartial = foundPatternPartial.replace("/", path.sep);
`

It should replace the slashes found in partials with correct path separators. But I haven't gotten around a PR :/

somatonic:
[...]

PR #675, opened 2017-08-11
Fix the verbose partial #559

gergan:
Addresses #559

Summary of changes:
We have the same problem as the reporter of #559 but using the handelbars implementation. I think the root of all problems is that almost everywhere the partial name currently is compared with subdir concatenated with '/' and filename.

subdir contains the OS native representation of the relative path, so it will work on linux but not on windows as it will contain backslashes and the rest will be generated with slashes.
I have found the property verbosePartial, which strangely enough is not used anywhere, but has the same value as the things against which pattern_assembler and pattern_registry are testing the partials.

So my solution was to 'normalize' the verbose partial with only forward slashes and use it consequently throughout the sources.

I will create a pull request for the Handelbars implementation (patternengine-node-handlebars), which than makes it work with the default include syntax.

The assumption for the two patches is that the templates are using only forward slashes for the inclusion, which is the only reasonable assumption in my opinion.


bmuenzenmeyer:
hello @gergan 

thanks for diving into this issue and looking into a fix - I will admit the verbose include syntax is not something I use often.

gergan:
hello @bmuenzenmeyer 

my bad have overlooked the contribution guidelines

bmuenzenmeyer:
💯This looks great - demonstrates a willingness to dive into the code and improve it.

PR #691, opened 2017-09-28
More helpful errors

phawxby:
Summary of changes:

Simple change to output the filename and error message of a failed parse to make the problem easier to find when there are multiple json files for a single pattern


phawxby:
Try duplicate keys. it definitely catches those

bmuenzenmeyer:
ugh, that didnt turn out as planned so i reverted

Issue #692, opened 2017-09-28
New `data_loader.js` module `loadFile` method loads the wrong file


phawxby:
I am using Pattern Lab Node `v2.11.1` on `Windows`, with Node `v8.4.0`, using the `Gulp` Edition.

##### Steps to Reproduce

Given this file structure
* hero
  * hero-left-short.json
  * hero-left-short,mustache
  * hero-left.json
  * hero-left.mustache

##### Expected Behavior

`hero-left.mustache` should build using `hero-left.json`

##### Actual Behavior

`hero-left.mustache` builds using `hero-left-short.json`

##### Example

Adding the following lines to `loadFile`
```javascript
    if (dataFilesFullPath.indexOf("\\hero-left*.") > -1) {
      console.log(dataFilesFullPath);
      console.log(dataFiles);
      console.log(dataFile);
    }
```

Results in the following console output
```shell
hero/hero-left*.{json,yml,yaml}
[ 'hero/hero-left-short.json',
  'hero/hero-left.json' ]
hero/hero-left-short.json
```

bmuenzenmeyer:
Nice catch! Sorry for the :bug:

Will check this out asap

PR #693, opened 2017-09-28
Glob fix to prevent loading the wrong file

phawxby:
Addresses: #692

Modify glob pattern to exclude the json files of other patterns


phawxby:
@bmuenzenmeyer any chance of a patch release so I can update?

phawxby:
:D Thanks!

Issue #684, opened 2017-09-18
Pin our dependencies


bmuenzenmeyer:
We should attempt to avoid problems like what happened at  js-beautify/issues/1247 in the future.

https://github.com/pattern-lab/patternlab-node/blob/master/package.json#L11

ms10398:
I want to work on this can i know what I have to do in this

bmuenzenmeyer:
@ms10398 basically we need to remove the carets "^" from our dependencies defined within `package.json` so that they don't update without our knowledge

ms10398:
Okay I will make a relevant PR soon. 
Thanks @bmuenzenmeyer

scarabcoder:
Is this still open?

bmuenzenmeyer:
No but plenty other items to choose from.  I have plans to add more next
week too
☆☆☆☆

PR #703, opened 2017-10-01
Pin all dependencies

ms10398:
Addresses https://github.com/pattern-lab/patternlab-node/issues/684

Summary of changes:

I have pinned all node dependencies


phawxby:
Why would you do this over a yarn.lock or a package-lock.json? Genuinely curious

bmuenzenmeyer:
@phawxby 

Appreciate the question - it's a good one.

Pattern Lab has always been a balancing act between approachability and power. One reason we still ship with Mustache as the default is how simple it is. This fact was relayed to me just today by a designer. Shipping with hawt new stuff is sometimes more to learn for people already harried by many other factors - and therefore creating a base experience is important.

* yarn is another layer of complexity. one that does curry many benefits, yes.
* npm 5 is required for a package-lock.json file - which I think we will get to eventually

Until support (which follows LTS) reaches the point where we can assume most users are on Node 8 (perhaps that's soon judging by https://github.com/nodejs/Release) I intentionally wanted to do the simplest thing that would make users a bit more safe (albeit potentially behind in deps)

phawxby:
So essentially it comes down to the users may not have Yarn and may not be running Node 8, version pinning allows for specific versioning with legacy support

ms10398:
Do I need to change anything in my PR

bmuenzenmeyer:
@ms10398 

> Do I need to change anything in my PR

You are good - the build is breaking due to something unrelated which I will look into.

jfroffice:
I don't think this was the best way to fix dependencies. 

You could have use `npm shrinkwrap` for this use case.

bmuenzenmeyer:
@jfroffice I think that's a fair criticism. Admittedly I know nothing about how shrinkwrap works. On the 3.X branch we will be landing soon, I will be implementing the package-lock.json concept anyways.

jfroffice:
`npm shrinkwrap` generates a __npm-shrinkwrap.json__ like __package-lock.json__.

it is used to freeze your dependencies on a working version.

So on, you can use `npm update` to increment your version using the semver notation and you don't have to touch the `package.json`. Except when you want to upgrade your dependencies.

>(and possibly re-add the carets?)

hopefully

Issue #688, opened 2017-09-27
Pattern link variable doesn't work in a pattern parameter


bradfrost:
I'd like to be able to use the [link variable](http://patternlab.io/docs/data-link-variable.html) in conjunction with a pattern parameter. So in essence, this:

```
{{> molecules-button(buttonText: "Complete Order", url : "link.pages-confirmation") }}
```

##### Expected Behavior
The link variable gets interpreted, and clicking on the button links through to the appropriate Pattern Lab pattern.

##### Actual Behavior
When the link variable is wrapped in quotes (as per the example above), it gets exported as a string as-is. When the link variable isn't wrapped in quotes, the build breaks.

##### Steps to Reproduce
See above code.

bradfrost:
Whoops sorry for the dupe!

bmuenzenmeyer:
No problem.  Your report was thorough so I kept this one. Thanks fur
continuing to report issues or wishes.

mwinkelm:
Sorry, that I comment on a closed comment, but I discovered a similar issue. Linking a page through parameters works fine, but linking a pseudo pattern page doesn't. In this case no link is generated at all.

In my case I have something like `{{>atoms-link-button:primary|expanded|large(cta: "Checkout", href: "{{link.pages-checkout~auth}}")}}`

In my checkout.mustache I have the following block:

```
{{#auth}}
  {{>templates-auth}}
{{/auth}}
{{#cart}}
  {{>templates-cart}}
{{/cart}}
```
And I use `auth` and `cart` to enable/disable content blocks in the pseudo pattern. The `atoms-link-button` lives in `templates-cart`.

I don't want to create an own page for each content block.

Is that simply not implemented or am I doing something wrong?

Thanks
Markus

PR #708, opened 2017-10-05
Add support for pattern parameter links

dagjaneiro:
Allow the use of link variables inside pattern parameters.

Addresses #688

Summary of changes:

The pattern parameters are run through `parse_data_links_specific` before rendering the pattern.


bmuenzenmeyer:
@dagjaneiro this is a great addition! thanks so much!

😍

PR #727, opened 2017-10-19
Removing redundant config.theme

EvanLovely:
What was happening is that the `theme` object from `patternlab-config.json` was getting put into `styleguide/data/patternlab-data.js` in both `config.theme` and as `theme`. I think it makes more sense to have settings declared in `config` be inside of that object. 

Relates to https://github.com/pattern-lab/styleguidekit-assets-default/issues/91


bmuenzenmeyer:
Considering how we are implementing https://github.com/pattern-lab/styleguidekit-assets-default/pull/93/ I am cool with this.

bmuenzenmeyer:
Changed base to `dev` per the contribution guidelines.

Issue #737, opened 2017-10-23
Post-install with starterkits, always clears source dir


mbulfair:
I am using Pattern Lab Node `v2.12.0` on `Windows `, with Node `v6.10.2`, using the `N/A ` Edition.

##### Expected Behavior
Installed starterkits not to clear source directory when installing new npm dependencies.

##### Actual Behavior
Every time a npm install/update is run, if a starter kit is there, source is always wiped.

##### Steps to Reproduce
Install any staterkit as a npm dependency 
Add a file anywhere in source
Run `npm install`

PR #738, opened 2017-10-23
patch fix for pattern-lab/patternlab-node/#737 starterkit postinstall

mbulfair:
Addresses #737 

Summary of changes:

Added new config flag for starterkit postinstall
Added a new variable to either act as current, or read from config

This is the same as the CLI version of passing --clean

### DIFF
--- a/core/scripts/postinstall.js
+++ b/core/scripts/postinstall.js
@@ -19,9 +19,12 @@ try {
   var starterkit_manager = new sm(config);
   var foundStarterkits = starterkit_manager.detect_starterkits();
 
+  //todo - find a better solution for postinstall cleaning of starterkits
+  var clean = (typeof config.starterkitPostInstallClean !== 'undefined') ? config.starterkitPostInstallClean : true;
+
   //todo - enhance to support multiple kits with prompt for each or all
   if (foundStarterkits && foundStarterkits.length > 0) {
-    starterkit_manager.load_starterkit(foundStarterkits[0], true);
+    starterkit_manager.load_starterkit(foundStarterkits[0], clean);
   } else {
     console.log('No starterkits found to automatically load.');
   }
--- a/patternlab-config.json
+++ b/patternlab-config.json
@@ -56,6 +56,7 @@
   "patternExportPatternPartials": [],
   "patternExportDirectory": "./pattern_exports/",
   "cacheBust": true,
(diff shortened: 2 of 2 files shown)


core/scripts/postinstall.js
```diff
@@ -19,9 +19,12 @@ try {
   var starterkit_manager = new sm(config);
   var foundStarterkits = starterkit_manager.detect_starterkits();
 
+  //todo - find a better solution for postinstall cleaning of starterkits
```
tburny: How could a better solution look like?

bmuenzenmeyer:
While I think this is a suitable workaround - I think a better approach would be to implement a starterkit feature similar to plugins where initial installation will add an entry to `patternlab-config.json` like:

**initial state**

``` javascript
"starterkits": []
```

**after installation, either from core, the CLI, or a human**

``` javascript
"starterkits": [
   {
      "name": "starterkit-mustache-demo",
      "installed": false // optional at this point
   }
]
```

**pattern lab runs**
**sees the entry and that it is not installed**

``` javascript
"starterkits": [
   {
      "name": "starterkit-mustache-demo",
      "installed": true
   }
]
```

doing this would still allow us to achieve similar functionality as today, where a user could ask pattern lab to delete the contents of `source/` if they really wanted to


``` javascript
"starterkits": [
   {
      "name": "starterkit-mustache-demo",
      "installed": false // optional at this point
      "deleteSource": true
   }
]
```

once installed, however, `deleteSource` would not take effect

``` javascript
"starterkits": [
   {
      "name": "starterkit-mustache-demo",
      "installed": true // set by pattern lab
      "deleteSource": true // has no effect, could warn about this
   }
]
```

Looking for feedback on this. I like that it mirrors the pattern that plugins has took and informs users in a declarative way what starterkits are installed

bradfrost:
> Looking for feedback on this. I like that it mirrors the pattern that plugins has took and informs users in a declarative way what starterkits are installed

@bmuenzenmeyer I think this approach makes sense to me.

mbulfair:
@bradfrost I agree with @bmuenzenmeyer approach.

raphaelokon:
I also agree with all of you. It makes a config also more portable and declarative. The PatternLab Node CLI also implements `install/enabling/disabling` of plugins so that functionality could be reused for starter kits as well I guess

Issue #749, opened 2017-11-14
Provide a mechanism to expose the default config file


bmuenzenmeyer:
Pattern Lab Node should provide a method to supply a consumer with the default `patternlab-config.json` file, https://github.com/pattern-lab/patternlab-node/blob/master/patternlab-config.json

Consumers can then mutate / save their own copy without having to rely on other copies of this file floating around editions / the CLI.

PR #750, opened 2017-11-14
Adding getDefaultConfig function to core

raphaelokon:
Addresses #749 

Summary of changes:
Adding a (static) function to the patternlab core to allow users to obtain a default config object


bmuenzenmeyer:
tested using 

``` js
const patternlabModule = require('@pattern-lab/patternlab-node');
const config = patternlabModule.getDefaultConfig();
const patternlab = require('@pattern-lab/patternlab-node')(config);

...
```

perfect 👍

PR #758, opened 2017-11-20
Adding an .npmrc with default configs

dinukadesilva:
This avoids the registry conflicts

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+registry=https://registry.npmjs.org/
+always-auth=false


bmuenzenmeyer:
Hey @dinukadesilva 

Can you provide more context as to why this is important for most users? Are you using Pattern Lab within a custom/private/corporate registry and this came up?  I am just trying to understand further.  The private registries I am familiar with will proxy to the public one if not found.

raphaelokon:
This is user land stuff and should not be committed IMHO. Parts of a per-project `.npmrc` are ignored anyways when published. It is intended to be there when working locally on a project.

dinukadesilva:
Actually only developers would benefit out of this change. https://github.com/pattern-lab/patternengine-node-mustache/issues/13

raphaelokon:
@dinukadesilva Thanks for the reply. This seems like a per user config to me. 
@bmuenzenmeyer What are your thoughts?

dinukadesilva:
It's not user specific. But it makes the project more configured and quickly installable to anyone.

raphaelokon:
It _can_ contain sensitive data like (`.env` files as well) when you login per project (I do not do that btw). Again this are just my two cents, but I do not know a single big FOSS project that maintain an .npmrc. There are actually tools that recommend against that (like RunKit) to avoid leaks.

dinukadesilva:
https://docs.npmjs.com/files/npmrc

bmuenzenmeyer:
Closing this - we've gotten by for this long without it and I don't quite understand the need. I develop pattern lab behind a corporate firewall with an npm proxy and have had no trouble to date.

PR #759, opened 2017-11-20
Pattern Lab Node 3.0.0 Alpha 6

### LINKED ISSUE #572: Patternlab should become a concrete class
I am using Pattern Lab Node `v2.7.0 (dev)` on `Linux`, with Node `v4.6.2`, using the `Grunt ` Edition.

Currently, the `Patternlab` "class" is just an object (`{}`)which is extended here and there. Here a "convention over configuration" pattern is applied: Properties of the PL object are added as required. 

This approach has the following disadvantages:
- The overall set of propierties the PL object can have is not defined thoroughly
- It is difficult to provide some sensible defaults in unit tests. This causes lots of DRY code which could be replaced by something like `Patternlab.createEmpty(additionalSettings)`
- It became and becomes more and more a god-object required everywhere

To overcome these problems I propose
- to create a class with a predefined set of fields, similar to the `Pattern` class. This also would probably improve overall modularity of the application.
- Replace some dependencies on the PL class with what is *really* required for a given method (improves testability and modularity). The goal is to un-god the object a bit
- See if we can move some methods to the PL object.

The structure would be as follows:
`PatternlabEngine` has a `Patternlab` object with the fields `engines`, `package`, `config`, `events`, `graph` (#540), `patterns`, `,  `subtypePatterns`,  `data`, `listitems`,  `patternSection`,  `patternSectionSubType`,  `header`,  `footer`,  `partials`,  `data.link` (for completeness). Also it has at least a method to load+check the configuration from a JSON file.
Probably major parts of `patternlab.js` could eventually be moved into there.
Additionally it might be wise to apply a strategy pattern to how patterns are compiles (compile all, incremental, whatever).

Opinions?

bmuenzenmeyer:
Closes #572 

Summary of changes: https://github.com/pattern-lab/patternlab-node/wiki/ChangeLog#v300-alpha6


bmuenzenmeyer:
letting CI go - too tired to watch

bmuenzenmeyer:
🎉 🎉 🎉 🎉 🎉  

Anyone that wants to get that code coverage back up, please do.

Issue #762, opened 2017-12-06
Assets being doubled in public when serving


danwhite85:
When using patternlab.serve all assets are doubled in the public directory.

Example:
![plerror](https://user-images.githubusercontent.com/636470/33685086-4ff2a9de-da85-11e7-9404-5abfa2d13720.PNG)

version: v3.0.0-alpha.5

bmuenzenmeyer:
Hi @danwhite85 

First off, thanks so much for giving the alpha a shot! It's quite a bit of a shake up, and admittedly a WIP in places.

I noticed the copying algorithm is pretty naive in places, and this recursion is uncalled for. I'll take a look at this as soon as available.

bradfrost:
@bmuenzenmeyer Hrm, is this my doing?

bmuenzenmeyer:
@bradfrost No - the alpha version of Pattern Lab Node 3.0 is doing all the asset copying itself - no grunt or gulp necessary, and it's a bit too recursive or globbing on too much it seems 🐛

danwhite85:
I've noticed now that it's not only assets. Everything in the styleguide folder is duplicated as well.

raphaelokon:
I am looking into it and already found the issue. I'll fix it and then push a PR.

PR #768, opened 2018-01-11
Fixes dir duplication when serving files

raphaelokon:
Addresses #762 

Summary of changes:
Removed the basename append in `addDir` watcher


bmuenzenmeyer:
Thanks for this! I will take a look after I stabilize some work within #769

Issue #766, opened 2018-01-10
All _meta files being run through the mustache engine


danwhite85:
All _meta files being run through the mustache engine.

The reason is that the meta patterns have incomplete data when being rendered in https://github.com/pattern-lab/patternlab-node/blob/dev/core/lib/object_factory.js#L105 (specifically the fileExtension in this case). 

Logging out the meta _00-head.njk file at the top of the render function results in:

    {
        relPath: '.',
        fileName: '.',
        subdir: '',
        fileExtension: '',
        name: '-.',
        jsonFileData: {},
        patternBaseName: '.',
        patternName: '.',
        patternGroup: '',
        patternType: '',
        patternSubGroup: '',
        patternSubType: '',
        flatPatternPath: '',
        patternLink: null,
        patternPartial: '-.',
        verbosePartial: '/.',
        isPattern: true,
        isFlatPattern: true,
        patternState: '',
        template: '',
        patternPartialCode: '',
        lineage: [],
        lineageIndex: [],
        lineageR: [],
        lineageRIndex: [],
        isPseudoPattern: false,
        order: 9007199254740991,
        engine: {
            engine: {
                name: 'mustache.js',
                version: '2.3.0',
                tags: [Array],
                clearCache: [Function: clearCache],
                parse: [Function: parse],
                render: [Function: render],
                to_html: [Function: to_html],
                escape: [Function: escapeHtml],
                Scanner: [Function: Scanner],
                Context: [Function: Context],
                Writer: [Function: Writer]
            },
            engineName: 'mustache',
            engineFileExtension: '.mustache',
            isAsync: true,
            expandPartials: true,
            findPartialsRE: /{{>\s*?([\w\-\.\/~]+)(?:\:[A-Za-z0-9-_|]+)?(?:(?:| )\(.*)?(?:\s*)?}}/g,
            findPartialsWithStyleModifiersRE: /{{>\s*([\w\-\.\/~]+)(?!\()(\:[\w\-\|]+)(\s*\([^\)]*\))?\s*}}/g,
            findPartialsWithPatternParametersRE: /{{>\s*([\w\-\.\/~]+)(\:[\w\-\|]+)?(\s*\([^\)]*\))\s*}}/g,
[...]

Issue #769, opened 2018-01-12
Re-work the Pattern Parameter algorithm


bmuenzenmeyer:
Async work has exposed what I believe to be a flaw in my pattern parameter algorithm. 

Consider the following set of patterns:

**paramParent.mustache**

``` mustache
{{> test-paramMiddle(class: "foo") }}
```

**paramParent.json**

``` json
{
  "url" : "link.test-foo"
}
```

**paramMiddle.mustache**

``` mustache
<div class="{{class}}">
  {{> test-link }}
</div>
```

**link.mustache**

``` mustache
<a href="{{url}}">Click Here for Foo Pattern</a>
```

### Expected Outcome

``` html
<div class="foo">
   <a href="../../patterns/test-foo/test-foo.rendered.html">Click Here for Foo Pattern</a>
</div>
```

### Actual Outcome

``` html
<div class="foo">
</div>
```

### Analysis

After troubleshooting for a while, I've called into question my decision to consume (i.e. immediately render) patterns when a pattern parameter is present. With async rendering within loops and recursion, this is just a nightmare.

What I am now pivoting to is a replacement of values only.  

So we see `class` patternParameter, we replace the `paramMiddle` partial to:

``` mustache
<div class="foo">
  {{> test-link }}
</div>
```

and then carry on with additional resolution. I think this will yield results more consistent with user' expectations.

bmuenzenmeyer:
Making good progress on this over on https://github.com/pattern-lab/patternlab-node/tree/pattern-param-rework

PR #770, opened 2018-01-13
Use patterns created by processMetaPattern for rendering

danwhite85:
Use patterns created by processMetaPattern for rendering (instead of empty ones).

Addresses #766 

Summary of changes:
1. Set patternlab.userHead and patternlab.userFoot to their full patterns instead of just the extendedTemplate.

2. Replace empty patterns in render calls with full patterns.

### DIFF
--- a/core/index.js
+++ b/core/index.js
@@ -195,14 +195,6 @@ const patternlab_module = function (config) {
         //cascade any patternStates
         lineage_hunter.cascade_pattern_states(patternlab);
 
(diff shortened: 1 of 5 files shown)


bmuenzenmeyer:
This looks like a great improvement. I will take a look as soon as I stabilize some work I am doing within #769

bmuenzenmeyer:
Thanks @danwhite85 !!!!!

PR #785, opened 2018-01-21
Render single pattern error handling

geoffp:
A couple of very simple changes that I needed to make error reporting work as needed in the React engine.

The idea is just to put the main `this.engine.renderPattern()` call into the promise chain so the `.catch()` will catch and report any errors that occur inside the call itself.


bmuenzenmeyer:
TT Geoff - we are going to investigate further if this is the best approach

bmuenzenmeyer:
@geoffp how's progress on the React engine? How can I help?

Issue #717, opened 2017-10-18
Copy a pattern's optional javascript file


bmuenzenmeyer:
Broken out of https://github.com/pattern-lab/patternlab-node/issues/706 as this is an enhancement.

Given a pattern `foo.mustache`
if a `foo.js` exists
we should copy it to the output location of `foo.mustache`

- [ ] consider how this will work or conflict with react engine

PR #798, opened 2018-02-15
Adding routine to copy js along pattern build

raphaelokon:
Addresses #717 

Summary of changes:

- Adding copy routine to copy pattern-specific JavaScript in place

### DIFF
--- a/core/index.js
+++ b/core/index.js
@@ -12,6 +12,8 @@
 
 const packageInfo = require('../package.json');
 
+const { concat } = require('lodash');
+const copy = require('recursive-copy');
 const path = require('path');
 const updateNotifier = require('update-notifier');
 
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
@raphaelokon this works pretty slick - nice work! 

This is the original output:

image

I've changed it slightly so the file name output is the same as one would include the partial itself:

image

This has the following benefits.

1. Muscle-memory for inclusion

``` html
<script src="atoms-brands-colors.js"></script>
```

2. Users can shortcut this by doing the following instead!

``` html
<script src="{{patternPartial}}.js"></script>
```

This works because I am renaming the file during copy, as you'll see in my commit, and then explicitly adding `patternPartial` to the data passed to render

The source files still follow the same convention (same name, different ext):

image
 
### BONUS ✨ 

By adding `patternPartial` to the data - we can support conditional CSS or (perhaps verbose JS) in the meta patterns

``` html
<body class="{{ bodyClass }} {{ patternPartial }}">
```

``` html
<script>if ('{{patternPartial}}' === 'pages-splash') { ... }</script>
```

though I dont know if the later is a good idea ⚔️

PR #809, opened 2018-03-01
Adding scored fuzzy matching

phawxby:
- Fixes a minor performance bug with excessive splitting in fuzzy matching
- Added scored fuzzy matching as a fallback

### DIFF
--- a/core/lib/get.js
+++ b/core/lib/get.js
@@ -1,38 +1,58 @@
 'use strict';
 
 const logger = require('./log');
+const _ = require('lodash');
 
 module.exports = function(partialName, patternlab) {
-  //look for exact partial matches
-  for (let i = 0; i < patternlab.patterns.length; i++) {
-    if (patternlab.patterns[i].patternPartial === partialName) {
-      return patternlab.patterns[i];
+  if (patternlab.patterns.length > 0) {
+    //look for exact partial matches
+    for (let i = 0; i < patternlab.patterns.length; i++) {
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
Hey @phawxby good to hear from you again!

Can you provide some more details as to what the issue is, the performance impact you are seeing, and how it is improved with this fix?

phawxby:
Say for example you want to segment things a bit.
atoms/filter/radio.mustache

Based on the current system it ignores that filter folder, so you could reference it using > atoms-radio or radio. Doing this allows referencing via just filter-radio or atoms-filter-radio. It gives more flexibility

phawxby:
@bmuenzenmeyer ?

bmuenzenmeyer:
@phawxby is this something you would still find value in? i am trying to work through all these PRs that I've neglected for too long now

phawxby:
Yes potentially. Although due to requirement changes we're currently looking at switching to a different atomic design system

bmuenzenmeyer:
> Based on the current system it ignores that filter folder, so you could reference it using > atoms-radio or radio. Doing this allows referencing via just filter-radio or atoms-filter-radio. It gives more flexibility

I am going to close this as not aligned with the future goals of Pattern Lab. Supporting custom resolvers outside the core capabilities of mustache increase tightly couple your patterns to Pattern Lab - making them less portable. 

I hope you understand and am sorry for keeping this open for so long. I hope you continue to find Pattern Lab of use but understand if you don't.

Issue #815, opened 2018-03-02
Monorepo Cleanup


bmuenzenmeyer:
Successor to https://github.com/pattern-lab/patternlab-node/issues/803

- [x] update contribution guidelines

For each converted repo:

- [ ] update readme to point to https://github.com/pattern-lab/patternlab-node
- [ ] [archive](https://help.github.com/articles/about-archiving-repositories/)
  - [ ] cli
  - [ ] core
  - [ ] edition-node-gulp https://github.com/pattern-lab/edition-node-gulp/issues/124
  - [ ] edition-node https://github.com/pattern-lab/edition-node/issues/6
  - [ ] engine-handlebars https://github.com/pattern-lab/patternengine-node-handlebars/issues/13
  - [ ] engine-liquid https://github.com/pattern-lab/patternengine-node-liquid/issues/1
  - [ ] engine-mustache https://github.com/pattern-lab/patternengine-node-mustache/issues/16
  - [ ] engine-react https://github.com/pattern-lab/patternengine-node-react/issues/6
  - [ ] engine-twig https://github.com/pattern-lab/patternengine-node-twig/issues/4
  - [ ] engine-underscore https://github.com/pattern-lab/patternengine-node-underscore/issues/4
  - [ ] live-server https://github.com/pattern-lab/live-server/issues/1
  - [ ] plugin-tab https://github.com/pattern-lab/plugin-node-tab/issues/25
  - [ ] uikit-workshop

The bulk of Pattern Lab links will be within `core/README`
- [ ] fix links
  - [ ] cli
  - [x] core
  - [x] edition-node-gulp 
  - [x] edition-node 
  - [x] engine-handlebars 
  - [x] engine-liquid 
  - [x] engine-mustache 
  - [x] engine-react 
  - [x] engine-twig
  - [x] engine-underscore 
  - [x] live-server
  - [x] plugin-tab  
  - [x] uikit-workshop

- [x] hoist devDependencies like prettier

- [x] Fix links within wiki
- [x] Fix links within patternlab.io
- [x] Fix core documentation generation (if needed)

- [x] update https://github.com/bmuenzenmeyer/patternlab-node-3.0-demo-vanilla
- [x] update https://github.com/bmuenzenmeyer/patternlab-node-3.0-demo-integration

raphaelokon:
Alos fix refrences to styleguidekit and assets to point to `uikit-workshop` in the default config.

bmuenzenmeyer:
[...]

PR #827, opened 2018-03-16
More cleanup

bmuenzenmeyer:
Part of #815 

Summary of changes:

* hoists common devDependencies so we are on the same version of eslint* and prettier
* excludes tests from code coverage


bmuenzenmeyer:
I expect coverage to go down after it was artificially inflated

PR #839, opened 2018-04-13
Test tree location proposal

geoffp:
This is a proposal to create a top-level directory, distinct from `packages`, to put pattern trees intended for various kinds of testing -- in my case the motivation was testing patterns for the React engine.

This also adds the newly freshened testing tree for React that is its reason for being. I could see this being a good place for the testing trees we use for file-based unit tests!

#### Why have this in the monorepo at all?
Well, because it *has* to depend on in-flight versions of the React engine, uikit, and core, and if it's outside the monorepo, even doing an `npm install` is janky since one then also has to write very loose version strings in `package.json`, or else face install errors (since we depend on unpublished versions of PL packages). Even then, you still have to `npm link` at least those three moving parts every time you need to regenerate your `node_modules`. If you're in the monorepo, Lerna does that all for you, and the whole thing is much less fragile.

#### But why, Geoff, why a separate directory?
Because these are not intended to be published, pretty much ever, and this avoids cluttering up the nice pristine `packages` directory with things that, well, won't wind up as actual packages. `packages` can remain the canonical directory of all that is to be published.

### DIFF
--- a/packages/development-edition-engine-react/package.json
+++ b/packages/development-edition-engine-react/package.json
@@ -7,7 +7,7 @@
   "dependencies": {
     "@pattern-lab/core": "^3.0.0-alpha.11",
     "@pattern-lab/engine-mustache": "^2.0.0-alpha.5",
-    "@pattern-lab/engine-react": "0.3.0",
+    "@pattern-lab/engine-react": "0.2.1-alpha.2",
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
@geoffp think this is good to go?

geoffp:
Looks good to me!

PR #840, opened 2018-04-13
Feature/uikit refactor

### LINKED ISSUE #811: Re-architect how ui's are built
As a user, I want to be able to generate multple UIs from a single pattern tree.

We will basically support a workflow first described in @bradfrost's article, [The Workshop and the Storefront](http://bradfrost.com/blog/post/the-workshop-and-the-storefront/).

Pattern Lab Node needs to change to accept N number of UIs to build. This issue will attempt to capture what is needed for that to be successful.

I think an acceptable way to think of how this would work is to start at the configuration point and move forward from there.

### Proposed Solution

**patternlab-config.json**

``` json
 "uikits": [
 {
   "name": "@pattern-lab/uikit-workshop",
   "outputDir": "workshop",
   "enabled": true,
   "excludedPatternStates": [],
   "excludedTags": []
 },
 {
   "name": "@pattern-lab/uikit-storefront",
   "outputDir": "storefront",
   "enabled": true,
   "excludedPatternStates": ["inprogress", "inreview"],
   "excludedTags": []
 },
]  
```

### High-Level Pseudocode `build()`

1. Build source patterns
1. `require()` all uikits by name
1. For each uikit in uikits...
   1. Build frontend...
       1. bail if not enabled via config or maybe a runtime flag
       1. exclude patterns as indicated by state and tag exclusions in the config
       1. output to `outputDir` as root - join to `paths.public`
   1. Copy assets into frontend

From what I can think about it right now, uikit build could be parallelized. Not sure if it's worth doing that, but may be neat!

### Questions

1. Would all uikits share the same `_meta` files? That would certainly be easiest, but may be limiting.

### Evaluation

:x: `index.js`

* `headPatternPromise` and `footPatternPromise` decomposed but only used within `ui_builder` (a good thing from a refactoring standpoint)

:x: `patternlab.js`

* The assigment of templates hanging off of `patternlabFiles` assumes a single uikit
   * `header`
   * `footer`
   * `patternSection`
   * `patternSectionSubType`

* `writePatternFiles()` assumes a single uikit

:x: `ui_builder.js`

* assumes a single uikit
* needs to filter out states and tags not wanted to output
 

:x: `exportData.js`

* assumes a single uikit
* might be a good extension point for custom logic within a uikit. use events?

:x: `copier.js`

* assumes a single uikit

:x: `pattern_graph.js`

* `exportToDot()` assumes a single uikit

✔️ `pattern_exporter.js`

* assumes a single location - but I think that is okay

_____

MISC

1. We need to well-document part of the UI-build process we have previously let slide if we want teams to adopt this approach (or even if we want to easily create `uikit-storefront`
   * the data schema available to a uikit at render time 
   * the expected templates a uikit provides
   * the file locations Pattern Lab will output after build

1. We need to update Ecosystem documentation to refer to uikits and the ability to have N of them

1. It's probably revisting in `meta` patterns are truly optional. We try to fall back to general header and footer but I need to test if this works in practice.

bmuenzenmeyer:
Closes #811 

Summary of changes:

**UIKits** are a new term in the [Pattern Lab Ecosystem](http://patternlab.io/docs/advanced-ecosystem-overview.html), attempting to evolve beyond the current `Styleguidekit` pattern which separated front-end templates from front-end assets like stylesheets and code. The existing `styleguidekit-assets-default` and `styleguidekit-mustache-default` have merged into `uikit-workshop` (more on that name in a bit). The hope is that by co-locating all front-end files in a single repository, users will be more capable of altering existing UIKits, making their own, even building multiple UIs from a single source of truth.

You heard that right - Pattern Lab 3.X will support multiple UI outputs from a single set of source patterns - implementing what @bradfrost coined [the workshop and the storefront](http://bradfrost.com/blog/post/the-workshop-and-the-storefront/). True to form, the default Pattern Lab experience is therefore the workshop.

`patternlab-config.json` has a new key:

``` diff
+  "uikits": [
+    {
+      "name": "uikit-workshop",
+      "outputDir": "",
+      "enabled": true,
+      "excludedPatternStates": ["inreview"],
+      "excludedTags": []
+    }
+  ]
```

`uikits` accepts and array of UIKit objects, shipping with the one above.
* `name`: the name of the UIKit
* `outputDir` where to output this UIKit relative to the current root. by leaving this empty we retain the existing behavior, outputting to `<project_rooter>/public`. If you had multiple UIKits, however, you would provided different values, such as:
``` json
"uikits": [
  {
    "name": "uikit-workshop",
    "outputDir": "workshop",
    ...
  },
  {
    "name": "uikit-storefront",
    "outputDir": "storefront",
    ...
  }
]
``` 

check the output:

image

* `enabled`: quickly turn on or off the building of this UIKit
* `excludedPatternStates`: tell Pattern Lab not to include patterns with these states in this UIKit's output
* `excludedPatternTags`: tell Pattern Lab not to include patterns with these tags in this UIKit's output
   * currently not supported

Other notes that coincide with this work:

* the [default `paths.source` object paths](https://github.com/pattern-lab/patternlab-node/pull/840/commits/a4961bd5d696a05fb516cdd951163b0f918d5e19) within `patternlab-config.json` are now relative to the current UIKit. See the [structure of uikit-workshop](https://github.com/pattern-lab/patternlab-node/tree/master/packages/uikit-workshop) for more info
* the [default `paths.public` object paths](https://github.com/pattern-lab/patternlab-node/pull/840/commits/812bab3659f504043e8b61b1dc1cdac71f248449) within `patternlab-config.json` are now relative to the current UIKit's `outputDir`. Absolute paths will no longer work. Someone could test putting an absolute path in a UIKit `outputDir` property and see what happens I suppose.
* `dependencyGraph.json` has moved to the project root rather than `public/` as we should only retain one

To aide in users building their own UIKits, I've also created [uikit-bare](https://github.com/pattern-lab/uikit-bare) outside this monorepo, as a thing to fork and change to your needs. The [uikit-bare wiki](https://github.com/pattern-lab/uikit-bare/wiki) attempts to document the API Pattern Lab expects from a UIKit and what data it provides a UIKit during build.  

#### Known Issues

* UIKit excluded tag support
* UIKit lineage still links to excluded patterns

#### Feedback Needed

This is a work in progress, so please provide feedback along the way.


bradfrost:
Just stopping in to say hello, as this looks fun and interesting!

PR #853, opened 2018-05-16
fix(watch): wire up serve and watch listeners correctly

### LINKED ISSUE #852: When serving, patterns are not rebuilt on save
When serving, patterns are not rebuilt on save. Using 3.0.0-alpha.13

##### Expected Behavior
Patterns would be rebuilt, the browser reloaded, and changes displayed
##### Actual Behavior
Browser reloads, but no changes displayed. 
##### Steps to Reproduce
Setup a clean instance of [edition-node ](https://github.com/pattern-lab/patternlab-node/tree/master/packages/edition-node). Create a pattern. Serve. Modify pattern.

bmuenzenmeyer:
Closes #852

Summary of changes:

Allows for more than one listener to be wired up to  `PATTERNLAB_PATTERN_CHANGE`. Curious if this will need to ever change should plugins or user behavior dictate it. If so, I think that we should be able to check for the presence of these specific handlers.


bmuenzenmeyer:
still need to account for build / watch / no serve option, in code, user testing, and unit tests

PR #918, opened 2018-08-03
Nunjucks loader to resolve Pattern Lab include paths

danwhite85:
Creates a custom Nunjucks loader to resolve Pattern Lab paths. 

This replaces the hacky search and replace method that exists currently. Also fixes an issue with including patterns that contain includes.


bmuenzenmeyer:
You can revert it ideally.  Lerna increments based on conventional
commits.

https://conventionalcommits.org/

danwhite85:
Reverted

PR #936, opened 2018-08-29
Set noCache = true on Nunjucks env, so restarts are not needed so often

danwhite85:
Prevents Nunjucks from caching templates, so a PL restart is not needed to see template changes.


bmuenzenmeyer:
@danwhite85 not sure whats up with the build - we can try adding https://github.com/pattern-lab/patternlab-node/commit/125b1117f881177cd72b7e1410c7f74b8ddb6562 or force merge and you can test your Nunjucks work locally. thoughts?

PR #945, opened 2018-09-14
fix: Prettier should only format .js files

### LINKED ISSUE #910: Prettier should only format .js files
In addition the the current [prettierignore](https://github.com/pattern-lab/patternlab-node/blob/master/.prettierignore), we [don't want prettier](https://github.com/pattern-lab/patternlab-node/pull/903#issuecomment-407456109) messing with `.json`, `.md` or `.scss` files.

Make sure this works:

* run via `npm run prettier`
* within any IDE integrations that honor `.prettierignore` and `.prettierrc`
* within the precommit hook

jbatley:
Closes #910 

Summary of changes: Updated .prettierignore and anded ignore path to npm prettier script

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -9,3 +9,6 @@ packages/core/scripts/api.handlebars
 packages/core/scripts/events.handlebars
 packages/core/test/files/annotations.js
 **/uikit-workshop/src/js/**/*
+*.json
+*.md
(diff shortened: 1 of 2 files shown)


bmuenzenmeyer:
Thanks for this! Sorry for the delay

PR #959, opened 2018-10-11
Feature/handlebars helpers

### LINKED ISSUE #958: How to write my own handlebars helper and use it in patternlab-node?
I am using Pattern Lab Node `v2.12.0` on `Windows`, with Node `v8.11.1`, using a `Gulp` Edition.

I am very confused how to actually write my own handlebars helper and use it in patternlab-node. I read there are patternlab plugins but is this the correct way to load a handlebars helper?

I hope someone can point me in the right direction. I also didn't find any information about plugins or whatsoever on the official documentation page.

Thanks in advance.

engelfrost:
Closes #958 

Summary of changes:
- Specify one or several sources for Handlebars helpers in `patternlab-config.json`. Structure of the config copied from `@pattern-lab/engine-react`. 
- Give `engine-handlebars` the ability to load helpers. 
- Demonstrate loading helpers in `development-edition-engine-handlebars`


bmuenzenmeyer:
@engelfrost thanks for this! it's support we really should have had sooner. 

Do you think this implementation is any superior to this one? It'd be nice to have some consistency
https://github.com/pattern-lab/patternlab-node/tree/dev/packages/engine-nunjucks#extending-the-nunjucks-instance

cc @danwhite85

engelfrost:
Which solution is superior depends on how Pattern Lab is supposed to function. I'll lay out the major differences, maybe you or someone else with a more holistic perspective can say which way is the "Pattern Lab way" of doing things. 

1. The major difference is that my solution specifies the source for helpers in the config, while the nunjucks engine requires a single specific file. My solution is more flexible, but requires more configuration. Seems like a typical convention vs configuration situation, so it's more a question of consistency and style. 

2. My solution requires the engine loader to call `usePatternLabConfig()`, like `engine-react`, while `engine-nunjucks` loads it straight from the JSON file. It seems reasonable to me that `core` is responsible for the configuration, who gets to use and possible manipulate it.

engelfrost:
After thinking about it more `usePatternLabConfig()` (or similar) should be the way to go, since you might want to pass PL another config.

engelfrost:
I've expanded my code to also look in a default location. If you think this is a good idea I could add the same configurability to engine-nunjucks.

bmuenzenmeyer:
@danwhite85 I am cool with these changes but want you to bless them in the nunjucks engine

engelfrost:
I've looked into implementing the same configurability into the Nunjucks engine, and there is a dependency issue in the `engine-nunjucks` code. The config is required in `engine_nunjucks.renderPattern()`, but the config is loaded from a hard-coded location instead of being injected. 

Is `usePatternLabConfig()` always called before `renderPattern()` is called? In that case the solution is trivial.

danwhite85:
I would definitely like to see a more structured approach than what I did with the Nunjucks engine. I didn't put a lot of thought in to it. Just did what was needed at the time. @engelfrost's solution looks promising. I will try to find some time to test the approach.

engelfrost:
@danwhite85 I've pushed my solution for `engine-nunjucks` to this PR in case it's of any help to you.

bmuenzenmeyer:
👏 

will try to review this this weekend. any other reviews would be outstanding. especially from you @danwhite85

danwhite85:
@engelfrost, this is great! Spent some time reviewing and testing. I'm happy with it if you are, @bmuenzenmeyer

engelfrost:
Slef critique: the Handlebars config option is engines.handlebars.helpers, but the Nunjucks option is engines.nunjucks.extensions. Is an engine agnostic option name preferable? Perhaps just "extend", since you need the wrapping function (not just the helper/extension)

mlnmln:
@dnwhte @engelfrost I'm very much interested in this PR landing. What is preventing it? How could I be of help?

engelfrost:
@dnwhte anything preventing the review from being considered complete?

dnwhte:
Looks good to me.

engelfrost:
Any chance of merging and releasing this soon?

mlnmln:
@dnwhte @engelfrost What't the status here? I'm starting to wonder about the maintenance model of PL node. Do you guys need more time / resources? Let me know if I could be of help.

bmuenzenmeyer:
> What't the status here? I'm starting to wonder about the maintenance model of PL node. Do you guys need more time / resources? Let me know if I could be of help.

The status here is that I went on hiatus for more than 8 months. I have limited time and always welcome contributions

bmuenzenmeyer:
1 conflict yet to be resolved

bmuenzenmeyer:
@engelfrost this is great work - sorry it took so long for me to prioritize it. I owe you an apology. Hoping to get into the swing of things more - and your help and support is always welcome

❤️

Issue #895, opened 2018-07-13
Installing Twig Engine on new project causes errors on npm run patternlab build


hdwebpros:
I am using Patternlab edition-node@1.0.0-alpha.13 on Windows, with Node 9.9.0, using a Vanilla Edition.

##### Expected Behavior
`npm run patternlab build` turns .twig files into public twig patterns without error

##### Actual Behavior
Seems like for each pattern you get:
`(node:11932) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.`
and then
```
{ Error: Command failed: php C:\websites\fail3\node_modules\exec-php\lib\php\cli.php -pC:\Users\RYANBO~1\AppData\Local\Temp\tmp-11932tjn5me1.m9qf.tmp -rC:\Users\RYANBO~1\AppData\Local\Temp\tmp-11932tus7wsp.18ge.tmp
PHP Fatal error:  Uncaught Twig_Error_Loader: The "..\..\..\..\source\_patterns" directory does not exist ("C:\websites\fail3\node_modules\node-twig\php\..\..\..\..\source\_patterns"). in C:\websites\fail3\node_modules\node-twig\php\vendor\twig\twig\lib\Twig\Loader\Filesystem.php:101
Stack trace:
#0 C:\websites\fail3\node_modules\node-twig\php\vendor\twig\twig\lib\Twig\Loader\Filesystem.php(82): Twig_Loader_Filesystem->addPath('..\\..\\..\\..\\sou...', '__main__')
#1 C:\websites\fail3\node_modules\node-twig\php\vendor\twig\twig\lib\Twig\Loader\Filesystem.php(40): Twig_Loader_Filesystem->setPaths(Array)
#2 C:\websites\fail3\node_modules\node-twig\php\Twig.php(80): Twig_Loader_Filesystem->__construct('..\\..\\..\\..\\sou...')
#3 C:\websites\fail3\node_modules\exec-php\lib\php\functions.php(28): render('02-pages\\00-gen...', Array)
#4 C:\websites\fail3\node_modules\exec-php\lib\php\cli.php(14): _exec_php_call_user_function('C:\\websites\\fai...', 'render', Array)
#5 {main}
  thrown in C:\websites\fail3\node_modules\node-twig\php\vendor\twig\twig\lib\Twig\Loader\Filesystem.php on line 101

    at ChildProcess.exithandler (child_process.js:273:12)
    at ChildProcess.emit (events.js:180:13)
[...]

PR #969, opened 2018-11-02
fix: correctly resolve paths

DanielRuf:
Partly fixes the issues described in https://github.com/pattern-lab/patternlab-node/issues/895


bmuenzenmeyer:
Ah I see now the separate branches. all PRs should target the default `dev` branch per https://github.com/pattern-lab/patternlab-node/blob/dev/.github/CONTRIBUTING.md#guidelines

PR #994, opened 2019-02-22
Feature/cli create command

JosefBredereck:
Summary of changes:

With the new CLI command, it's possible to create pattern files. It's possible to choose which files should be created and in which way they will be created.


JosefBredereck:
No stale bot you won't close this one

PR #1027, opened 2019-06-27
Removing padding that causes horizontal scrollbar

shaal:
You probably cannot see it easily on a mac (that hides scrollbars...), but there's a horizontal scrollbar as you can see in this screenshot:
image

Closes #

Summary of changes:


bradfrost:
Awesome, thanks so much for this @shaal!

PR #1035, opened 2019-08-09
Add module export to exportData

jryanconklin:
Closes # No-Issue

Summary of changes:

Per convo with @sghoweri on gitter, this adds a `module.exports` to exportData as the JSON version is removed. This allows us to have automated tests to pattern lab for accessibility, backstop, etc.


sghoweri:
I'll dig deeper into this later but just as an aside: we'll want to double check that all the data PL Node has available here is getting exported -- just in case other folks might need to pull in / wire up this data to other things as well 😉

For reference, we already had to patch something similar for [Pattern Lab PHP](https://github.com/bolt-design-system/bolt/blob/master/docs-site/.patches/exported-data.patch). 

Besides the automated testing use cases mentioned by @jryanconklin, there's other great uses for this data too (like our [component status board](https://boltdesignsystem.com/pattern-lab/?p=components-overview))!

sghoweri:
@jryanconklin FYI, I'm taking a look at this right now!

sghoweri:
Hmm looks like we might have one missing piece here that we need to solve...

I'm seeing a JS error after pulling in this update to test: image

^ besides transpiling this file with Babel (which is probably overkill), another option could be to output a browser-friendly version of this data file + a Node.js friendly version of this JS (at least as an interim solution).

sghoweri:
Update: Sorry, it looks like that exporting a 2nd file with the module.export addition was actually the solution we ended up using when we patched this same overall thing to PL PHP on our end.

Here's the interim update I ended up making to avoid conflicts between the two different JS file formats:
```
// add module.export to the Nodejs-specific file generated.
  const exportedOutput =
    output +
    'module.exports = { config, ishControls, navItems, patternPaths, viewAllPaths, plugins, defaultShowPatternInfo, defaultPattern, comments };';

  //write all output to patternlab-data
  _.each(patternlab.uikits, uikit => {
    fs.outputFileSync(
      path.resolve(
        path.join(process.cwd(), uikit.outputDir, paths.public.data),
        'patternlab-data.js'
      ),
      output
    );
    fs.outputFileSync(
      path.resolve(
        path.join(process.cwd(), uikit.outputDir, paths.public.data),
        'patternlab-data.cjs.js'
      ),
      exportedOutput
    );
  });
```

@jryanconklin can you test this updated code snippet out on your end (pointing to the `pattern-data.cjs.js` file output for your Node.js scripts) to see if this works for ya?

jryanconklin:
@sghoweri, awesome! Thank you! I’ll take a stab at testing this first thing tomorrow.

jryanconklin:
@sghoweri Confirmed the above mostly works. I had to leave `comments` off `module.exports` for it to work. I'll update this PR leaving `comments` off, but this fixed the reference error and unblocks some work I'm doing. THANK YOU.

PR #1038, opened 2019-08-23
feat(starterkits): add starterkit-handlebars-demo

bmuenzenmeyer:
@bradfrost's starterkit-handlebars-demo is not showing up during init

image

this used to be dynamic... but at some point i do not recall appears to be hardcoded.

### Testing

```
cd to/monorepo/packages/cli
npm link
```

then

```
mkdir pl-test-1038
cd pl-test-1038
npm init -y
npm install @pattern-lab/cli
npm link @pattern-lab/cli
npx patternlab init
```

image

image

image

though these choices are made - the actual code (such as handlebars as the default wont likely be in place until this is released)

### DIFF
--- a/packages/cli/bin/inquiries/edition.js
+++ b/packages/cli/bin/inquiries/edition.js
@@ -15,19 +15,15 @@ const editionSetup = [
     message: 'Which edition do you want to use (defaults to edition-node)?',
     choices: [
       {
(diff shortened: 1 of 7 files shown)


bmuenzenmeyer:
looks like there is custom logic in `install-edition` - why it doesnt just trust the contents and perform a large copy I am no longer sure...

PR #1059, opened 2019-09-18
Feature update dependencies

JosefBredereck:
Summary of changes:
Updated dependencies of different packages while they have npm audits for lodash, tap and even a critical with the handlebars package.

### DIFF
--- a/packages/live-server/package.json
+++ b/packages/live-server/package.json
@@ -12,14 +12,14 @@
   ],
   "author": "Tapio Vierros",
   "dependencies": {
-    "chokidar": "1.6.0",
+    "chokidar": "3.1.0",
     "colors": "1.3.0",
     "connect": "3.6.5",
     "cors": "2.8.4",
     "event-stream": "3.3.4",
     "faye-websocket": "0.11.x",
     "http-auth": "3.2.x",
-    "morgan": "1.6.1",
+    "morgan": "1.9.1",
     "object-assign": "4.1.1",
     "opn": "5.3.0",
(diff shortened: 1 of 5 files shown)


ringods:
@sghoweri or @bmuenzenmeyer is there anyone else who could process this PR and create a new release of the packages?

bmuenzenmeyer:
i can test this tomorrow

packages/live-server/package.json
```diff
@@ -12,14 +12,14 @@
   ],
   "author": "Tapio Vierros",
   "dependencies": {
-    "chokidar": "1.6.0",
+    "chokidar": "3.1.0",
```
bmuenzenmeyer: why upgrade this here but not in core? I am a big fan of using tools like https://www.npmjs.com/package/npm-check-updates to perform actions like this across the monorepo

bmuenzenmeyer:
tested this out - thanks for taking the time to improve the codebase. Will uptick that chokidar dependency while I am here

bmuenzenmeyer:
updating chokidar in core seems to have broken live-reload. confirming

packages/live-server/package.json
JosefBredereck: Hey @bmuenzenmeyer I walked my way through the directories and it seems like I missed the core with chokidar. Thanks for the note and providing a link to check the monorepro a better way.

bmuenzenmeyer: no problem! turns out it was good to omit for now anyways, see #1069 😊

PR #1066, opened 2019-10-04
🔌

### LINKED ISSUE #1021: Core should have a working plugin system
The current [documented procedure](https://github.com/pattern-lab/patternlab-node/wiki/Installing-Plugins) for installing plugins has two steps: 1) install plugin npm package 2) run Pattern Lab CLI install command.

The documentation seems to be out of date, but even when successfully running the CLI install command the plugin does not seem to run, and nothing seems to be added to `patternlab-config.json`.

Looking at core code there seems to be an additional plugin discovery mechanism, in which core looks for packages beginning with `plugin-` and tries to enable them. Puzzlingly, this discovery mecanism only seems to run if there is a `plugins` section in `patternlab-config.json`.

Manually adding a `plugins` section causes the discovery mechanism to run, but searching for packages beginning with `plugin-` can result in non-PL plugin packages to be treated as PL plugins. Babel for example has several packages beginning with `plugin-`.

It also seems that engines are discovered in a similar way, based on the name of the package.

I would suggest using more specific prefixes for package name based discovery. A better alternative to name based discovery might be to look for a section in a package's `package.json` identifying the package as a Pattern Lab plugin, engine etc.

There also seem to be some other issues related to the plugin system:

- [Investigate improvements to plugin system #812](https://github.com/pattern-lab/patternlab-node/issues/812)
- [[core] Remove plugin installation code #872](https://github.com/pattern-lab/patternlab-node/issues/872)
- [[cli] plugin installation not working #873](https://github.com/pattern-lab/patternlab-node/issues/873)
- [Installing a plugin results in an error when running patternlab. #916](https://github.com/pattern-lab/patternlab-node/issues/916)

In my tests I found especially [this pull request](https://github.com/pattern-lab/patternlab-node/pull/998) to be crucial to get plugins working.

bmuenzenmeyer:
Closes #1021 

Summary of changes:

adds new async callback system to plugins

- [x] PATTERNLAB_BUILD_START
- [x] PATTERNLAB_PATTERN_ITERATION_END
- [x] PATTERNLAB_PATTERN_BEFORE_DATA_MERGE
- [x] PATTERNLAB_PATTERN_WRITE_BEGIN
- [x] PATTERNLAB_PATTERN_WRITE_END
- [x] PATTERNLAB_GLOBAL_CHANGE
- [x] PATTERNLAB_PATTERN_CHANGE


bmuenzenmeyer:
will be working off this draft PR until stable

PR #1093, opened 2019-11-07
Feature/1092  twig namespaces

### LINKED ISSUE #1092: Engine Twig PHP isn't very flexible with namespaces
I am using Pattern Lab Node v5.1.0 and engine-twig-php v5.1.0 on Mac, with Node v11.2.0, using edition-twig (php engine).

##### Expected Behavior

I should be able to use a twig namespace that's not a directory name and includes / embeds etc should work.

##### Actual Behavior

When using a twig namespace that's not close to atomic naming conventions (Atoms, Molecules, etc), included partials aren't resolved correctly.

##### Steps to Reproduce

Within `patternlab-config.json` add a new namespace with an ID that's not similar to a directory:
```        {
          "id": "some_thing",
          "recursive": true,
          "paths": [
            "source/_patterns/05-test"
          ]
        },
```
Within another pattern twig file, include a new component using this namespace:
```
{% include "@some_thing/00-test.twig" %}
```

You should see the warning:

> Could not find pattern referenced with partial syntax @some_thing/00-test.twig.
> This can occur when a pattern was renamed, moved, or no longer exists but it still referenced within a different template or within data as a link.

I recognize this is a super specific bug that may only impact me and my team so I've got a PR coming. 😄

zhawkins:
Currently we're finding the twig partial folder by splitting the path based on the namespace ID (i.e. Atoms). This works well if we're using the atomic naming convention but not so much for others. At work we use Drupal theme names as our twig namespaces so this breaks in new and wondrous ways for us.

This PR updates how the `fullFolderPath` is generated. Instead of splitting on namespace ID we're splitting on the resolved namespace path. This allows us to leave the ID out of it while still figuring our the correct partial folder.

## Before:

`tempPartial = '/Users/foo/Sites/pattern-lab/source/_patterns/05-test/00-test.twig'`

`selectedNamespace[0].id = 'some_thing'`

```              
const fullFolderPath = `${
  tempPartial.split(selectedNamespace[0].id)[0]
}${selectedNamespace[0].id}`;
```

`fullFolderPath = '/Users/foo/Sites/pattern-lab/source/_patterns/05-test/00-test.twigsome_thing'`

## After:

`tempPartial = '/Users/foo/Sites/pattern-lab/source/_patterns/05-test/00-test.twig'`

`selectedNamespace[0].paths[index] = 'source/_patterns/05-test'`

```              
const fullFolderPath = `${
  tempPartial.split(selectedNamespace[0].paths[index])[0]
}${selectedNamespace[0].paths[index]}`;
```

`fullFolderPath = '/Users/foo/Sites/pattern-lab/source/_patterns/05-test'`

Closes #1092


zhawkins:
@sghoweri , I believe you worked on this code at some point so maybe you can help me figure out if there's a better way to do this. 👋

bmuenzenmeyer:
thank you @zhawkins and @mariohernandez for the fix and verification - it goes a long way toward helping make PL the cross-platform / cross-language thing we rely on

PR #1106, opened 2019-11-27
chore: UIKit - referencing 'html-webpack-plugin' from NPM

mfranzke:
As github might be inaccessible / blocked on some environments (for whatever/security reasons), it's probably better to reference this dependency from NPM instead of Github as well, and solve the problem of referencing to the beta version by its exact upper version number.

Closes #

Summary of changes:
"html-webpack-plugin" dependency is being retrieved via NPM instead of Github.

### DIFF
--- a/packages/uikit-workshop/package.json
+++ b/packages/uikit-workshop/package.json
@@ -91,7 +91,7 @@
     "hogan.js": "^3.0.2",
     "htm": "^1.0.1",
     "html-loader": "^0.5.5",
-    "html-webpack-plugin": "github:jantimon/html-webpack-plugin#webpack-4",
+    "html-webpack-plugin": "^4.0.0-beta.11",
     "iframe-resizer": "^3.6.5",
     "lit-element": "^2.2.1",
     "lit-html": "^1.1.2",


bmuenzenmeyer:
great point @mfranzke ! 

@sghoweri was there any particular reason we targeted the git url previously? The github changelog and releases vs npm results are a bit confusing

image

sghoweri:
> great point @mfranzke !
> 
> @sghoweri was there any particular reason we targeted the git url previously? The github changelog and releases vs npm results are a bit confusing
> 
> image

@bmuenzenmeyer if I’m remembering this correctly, the original reason for pointing to that Github repo branch (vs NPM) had to do with major Webpack v4-related issues that the available NPM versions published suffered from but the Github branch had already fixed.

It looks as if there’s been an uptick in recent activity + new betas published so this workaround might not be needed anymore! 

Let me double check that this still works with some of the performance related logic in the Webpack build (ex. generating Critical CSS); would love to swap back to the NPM version (and even perhaps move this to being a devDependency).

Thanks guys!

PR #1137, opened 2020-03-16
Allow uikit plugins to be used in multiple uikit definitions in patternlab-config.json.

### LINKED ISSUE #1136: Allow uikit plugins to be used across multiple uikit definitions
In patternlab-config.json it is possible to define multiple uikits with different configurations. As far as I understand, the idea is to have a workshop and a storefront uikit showing diffrent patterns based on the exclude configuration. 
Since there is no dedicated storefront uikit at the moment, it would be nice if you could just use iukit-workshop to generate the patternlab frontend. 

##### Expected Behavior
When I define multiple uikits with the same name, only one frontend is generated properly

##### Actual Behavior
Only the last one will work, since the config will override itself.

##### Steps to Reproduce
Put the following configuration into your patternlab-config.json
`"uikits": [
    {
      "name": "uikit-workshop",
      "outputDir": "workshop",
      "enabled": true,
      "excludedPatternStates": [],
      "excludedTags": []
    },
    {
      "name": "uikit-workshop",
      "outputDir": "storefront",
      "enabled": true,
      "excludedPatternStates": [
        "inprogress"
      ],
      "excludedTags": []
    }]`

##### Proposed solution
In order to be backwards-compatible, I suggest adding an optional key "id" to the uikit settings, which will be used to collect the configuration instead of the name. If the id does not exists, use the name as fallback.

Hydraner:
Closes #1136

Summary of changes:
- Adds an optional id parameter to uikit definitions in patternlab-config.json
- When collecting the uikits data, each over the config.json uikits instead of the available plugins


packages/core/src/lib/loaduikits.js
```diff
@@ -37,35 +37,42 @@ const readModuleFile = (kit, subPath) => {
...
 
-  const uikits = findModules(nodeModulesPath, isUIKitModule); // [1]
+  const uikitModules = findModules(nodeModulesPath, isUIKitModule); // [1]
 
-  uikits.forEach(kit => {
-    const configEntry = _.find(_.filter(patternlab.config.uikits, 'enabled'), {
-      name: `uikit-${kit.name}`,
+  _.filter(patternlab.config.uikits, 'enabled').forEach(uikit => {
+    const kit = _.find(uikitModules, {
+      name: uikit.name.replace('uikit-', ''),
```
JosefBredereck: Could you please resolve the merge conflict at this position, resulting from dev merge?

mfranzke:
@hydraner do you have a chance to review the comment by @josefbredereck?

ringods:
I would like another review of this implementation. In #1225 I also have to touch the loading of UI kits and plugins. In the context of Yarn v2 workspaces (monorepo), there isn't a `node_modules` folder in every package/workspace, but only at the top level. To correctly load dependencies, one has to `require` the package.

My suggestion for this PR is to use `name` for what you defined as `id`, but add `package` as the full name of package dependency, including scope. For example:

```json
  "uikits": [
    {
      "name": "uikit-workshop",
      "package": "@pattern-lab/uikit-workshop",
      "outputDir": "",
      "enabled": true,
      "excludedPatternStates": [],
      "excludedTags": []
    }
  ]
```

For backwards compatibility, if `package` is not defined, I would by default prefix the `name` property with the scope `@pattern-lab/`.

In the end, this would mean that you can use the same package multiple times, but under a different uikit name.

ringods:
In https://github.com/pattern-lab/patternlab-node/pull/1225/commits/ceccfd8814b638558b5f53446eb6625162a29c49, I have implemented the same in my branch. You can close this PR in favor of mine.

ringods:
@JosefBredereck can you please review my work in #1225. As mentioned before, this PR will conflict with my work which is nearing completion. There are a few more cases where manual `node_modules` calculation needs to be replaced with proper `require.resolve` before I can submit it fully for integration.

ringods:
@Hydraner are you ok with closing this PR in favor of #1225 which is just merged? There is only a difference in the names of the properties we both chose. You chose `name` and `id` and I chose `name` and `package`. Via my work, an NPM `package` can be used multiple times but with a different `name`. 

There is some backwards compatibility included when a `package` property is not provided in the uikit configuration, that the `name` will be tested with a few combinations to calculate a package name, combined with the proper deprecation warnings.

ringods:
This is covered by #1225 (which was reverted and redone in #1246). Documentation is added to the `uikits` section of the configuration:

https://patternlab.io/docs/editing-the-configuration-options/#heading-uikits

PR #1138, opened 2020-03-16
Introduce the ability to configure the nesting level of the pattern directory hierarchy.

### LINKED ISSUE #633: Ignore files nested 2 levels deep
Hello Team,
I put folders with assets in each of my pattern subtype folders. Those assets are js files, partials or whatever you want to put in there. 

File structure is like this:
```
//works fine
pattern/subtype/pattern-file.mustache
pattern/subtype/pattern-file.js

//these will trigger the 2 level warning
pattern/subtype/sources
pattern/subtype/sources/some-file.js
pattern/subtype/sources/_some-pattern.mustache
```

But I get a warning for those nested folders:

> "A pattern file [...] was found greater than 2 levels deep from [...]"
> It's strongly suggested to not deviate from the following structure under _patterns/
> While Pattern Lab may still function, assets may 404 and frontend links may break. Consider yourself warned.

The warning suggests that patternlab at least tries to process those files. I have concerns, that my files have side effects on the generating part of patternlab. You state yourself that it's strongly not suggested. So can we please just ignore files if too deep? 

This would save us from that huge list of warnings and ensure that no file will ever take part in the pattern processing.

What do you think? Thanks!

Hydraner:
Closes #633

Summary of changes:
-  Makes the depth of the directory hierarchy configurable
 - Introduces new "maxNestingLevel" option for patternlab-config.json

### DIFF
--- a/packages/core/src/lib/object_factory.js
+++ b/packages/core/src/lib/object_factory.js
@@ -32,17 +32,18 @@ const Pattern = function(relPath, data, patternlab) {
   info.dir = info.hasDir ? pathObj.dir.split(path.sep).pop() : '';
   info.dirLevel = pathObj.dir.split(path.sep).length;
 
+  this.maxNestingLevel = this.getMaxNestingLevel(patternlab);
   this.relPath = path.normalize(relPath); // '00-atoms/00-global/00-colors.mustache'
   this.fileName = pathObj.name; // '00-colors'
   this.subdir = pathObj.dir; // '00-atoms/00-global'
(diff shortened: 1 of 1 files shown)


JosefBredereck:
I would like to close this one based on the bigger changes in another ticket. What is your opinion @bmuenzenmeyer

bmuenzenmeyer:
This configurability is cool but has been an absolute nightmare to support. I fear that adding more and more layers is only going to make for a bigger house of cards to eventually fall. 

There were proposals a long time back to support a custom nav structure that users could wholly define. I feel like that's a potentially better approach - but not something I have bandwidth to explore or support.

PR #1140, opened 2020-03-17
Implement exclude tags feature.

### LINKED ISSUE #844: UIKit excluded tag support
As part of #811 and #840 we hoped to deliver `excludedTags` support within a UIKit. This is certainly possible, but I wanted to release early with state support first.

If someone wants to take on this work it would look similar to https://github.com/pattern-lab/patternlab-node/pull/840/commits/87c9d0d2c315d9f6060df70fc41a7f4a9a5e9f9d

* add this filter to `uikitExcludePattern`
* add tests

### LINKED ISSUE #1079: implement config option "excludedPatternTags"
I am using Pattern Lab Node `v3.0.4` on `Windows`, with Node `v12.12.0`, using a `Gulp` Edition.

##### Issue Description
The option `excludedPatternTags` in the `patterlab-config.json` so far has no effect at all.

The function is [described here](https://patternlab.io/docs/advanced-config-options.html) very briefly.
I assume the behavior is supposed to be the same as the `excludedPatternStates`:
- states are defined in [a patterns documentation file](https://patternlab.io/docs/pattern-documenting.html) like `state: incomplete` 
- `tags` has to be an array, like `tags: [new, relaunch, dev]`

##### Expected Behavior
Adding the option `excludedPatternTags` in the `patterlab-config.json` has an effect on the output of a build process. 
As with the `excludedPatternStates` pages should be completely excluded from a build if one of the by documentation file assigned tags matches one of the `excludedPatternTags`.

##### Actual Behavior
Using the option `excludedPatternTags` has no effect

##### Steps to Reproduce
Use the option `excludedPatternTags`, build a lab

Hydraner:
Closes #844
Closes #1079

Summary of changes:
- Extends uikitExcludePattern to also check for tags
- Tests for excludeTags config


bmuenzenmeyer:
Excited to see a feature long-promised add to the project!

packages/core/src/lib/uikitExcludePattern.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+var _ = require('lodash');
```
JosefBredereck: Eslint will throw an error on this line because var is not allowed.

```suggestion
const _ = require('lodash');
```

packages/core/test/uikitExcludePattern_tests.js
```diff
@@ -46,6 +46,53 @@ tap.test(
     //act
     const result = uikitExcludePattern(pattern, uikit);
 
+    //assert
+    test.true(result);
+    test.end();
+  }
```
JosefBredereck, line 52: Missing bracket leads to error while running tests

```suggestion
  }
);
```

packages/core/src/lib/uikitExcludePattern.js
```diff
@@ -1,7 +1,11 @@
...
+
 const uikitExcludePattern = (pattern, uikit) => {
   const state = pattern.patternState;
-  return uikit.excludedPatternStates.includes(state);
+  const tags = pattern.tags;
+
+  return uikit.excludedPatternStates.includes(state) || _.intersection(uikit.excludedTags, tags).length > 0;
```
JosefBredereck: `uikit.excludedPatternStates` is empty in one test, the same could happen with a `patternlab-config`. If we use lodash instead, the error will be catched and no `undifined` will occur.

```suggestion
  return (
    _.includes(uikit.excludedPatternStates, state) ||
    _.intersection(uikit.excludedTags, tags).length > 0
  );
```

packages/core/src/lib/uikitExcludePattern.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+var _ = require('lodash');
+
 const uikitExcludePattern = (pattern, uikit) => {
   const state = pattern.patternState;
-  return uikit.excludedPatternStates.includes(state);
+  const tags = pattern.tags;
```
JosefBredereck: In markdown, you could add one single tag by just providing something like the following. 
```
---
tags: foo-bar
---
```

This will result in only having a string in `pattern.tags` and results in intersection returning false. My suggestion is to check if the pattern tags are already an array and if not the string will be included in an array.

```suggestion
  const tags = _.isArray(pattern.tags) ? pattern.tags : [pattern.tags];
```

sghoweri:
@JosefBredereck @Hydraner are we good to go on this?

JosefBredereck:
As @bmuenzenmeyer accepted the review corrections it is ready.

Issue #1134, opened 2020-03-06
Since 5.7.0, PL menu structure breaks if filename and subfolder name match


coreylafferty:
I am using Pattern Lab Node `v5.7.1` on `Windows`, with Node `v10.15.1`.

##### Expected Behavior
Folder names and file names should not break the menu if they happen to be the same.
 
##### Actual Behavior
Prior to 5.7.0, the menu worked as expected when subfolder names and file names matched.  Since 5.7.0, menu functionality breaks when this is the case.  Clicking on submenu items will display the wrong items or no item at all.

@bmuenzenmeyer @david-moldenhauer I believe this is due to the changes from this PR, specifically these lines: 
https://github.com/pattern-lab/patternlab-node/pull/1016/files#diff-811dc08dbb9437f8b472a21f0c9bd66eR38-R40

##### Steps to Reproduce

Install a fresh copy of PL and create the following files/folders in /source/_patterns/:
```
02-organisms/broken/broken.twig
02-organisms/broken/broken-item1/broken-item1.twig
02-organisms/broken/broken-item2/broken-item2.twig
```
Open PL in the browser and click the arrow next to the 'Broken' menu under 'Organisms' then click on the first child (Broken Item2).  Note that it will actually show the wrong item and mark the wrong item as active in the menu.  

The issue here seems to be that the subfolder name (e.g., broken-item1) and the filename in it (broken-item1.twig) are identical (ignoring the file extension). 

To contrast, create the following files/folders in /source/_patterns/ to see an almost identical situation, but where everything works as expected.  The only difference here is that the subfolder file name (e.g., working1) does NOT match the filename in it (e.g., working-item1).  
```
02-organisms/working/working.twig
02-organisms/working/working1/working-item1.twig
02-organisms/working/working2/working-item2.twig
```

ringods:
Have you tried the pure JS twig engine from `@pattern-lab/engine-twig` lately? It got a big refresh in version 5.8.0 by using a better library under the hood.

coreylafferty:
Thanks, @ringods.  I think this is more tied to the PL navigation creation.  This issue appears to be fixed by PR #1143, which appears to be close to being accepted.

mfranzke:
Fight the bot

JosefBredereck:
Fixed with #1143 in [V5.12.0](https://www.npmjs.com/package/@pattern-lab/core)

Issue #714, opened 2017-10-17
Support view all pages in pattern linking


bmuenzenmeyer:
Implement https://github.com/pattern-lab/the-spec/issues/35

Pattern Lab exports `patternPaths` and `viewAllPaths` separately to the UI within [ui_builder](https://github.com/pattern-lab/patternlab-node/blob/master/core/lib/ui_builder.js#L555-L591), since they've never had to co-mingle prior to this spec enhancement.

Then, when patterns are linked, I believe they are checked only against `patternPaths` to resolve the location.

The first thing I'd try would be to concatenate `viewAllPaths` into the `patternPaths` export and then see if a link like `<a href="{{ link.viewall-atoms-all }}">Go to all Atoms</a>` works (assuming you have a Pattern Type called atoms).

If that doesn't work, more tech design needs to occur.

JosefBredereck:
Fixed with #1143 in [V5.12.0](https://www.npmjs.com/package/@pattern-lab/core)

Issue #660, opened 2017-05-09
Unable to generate "View All" pages on flat patterns


ThePeach:
I am using Pattern Lab Node `v2.9.2` on `Mac`, with Node `v7.9.0`, using the `Gulp` Edition.
I have a category directory with a flat list of mustache files in it, i.e. without sub-category folders.

##### Expected Behavior
I'm expecting the "View All" page to be generated for that directory

##### Actual Behavior
No "View All" page is being generated

##### Analysis
Apparently (after chatting with @bmuenzenmeyer) this is _by design_.

It would be indeed nice if it was possible to control the generation of "View All" pages programatically and granularly on a per-directory way, e.g. with something as simple as a dot-file in the directory to enable its generation, and/or with a global "default" option that won't break backward compatibility.

JosefBredereck:
Fixed with #1143 in [V5.12.0](https://www.npmjs.com/package/@pattern-lab/core)

PR #1143, opened 2020-04-12
Update pattern discovery and fix "view all" issues

JosefBredereck:
### Closes #

| Issue | Title |
|---|---|
| #890 | [uikit-workshop] View All for subtypes is broken |
| #1134 | Since 5.7.0, PL menu structure breaks if filename and subfolder name match |
| #633 | Ignore files nested 2 levels deep |
| #660 | Unable to generate "View All" pages on flat patterns |
| #714 | Support view all pages in pattern linking |
| #1016 | (errors resulting from) Ensure that patterns in directories greater than 2 levels deep from source are sorted to the next possible parent directory. |

### Replaces #

| Issue | Title |
|---|---|
| #1138 | Introduce the ability to configure the nesting level of the pattern directory hierarchy |

#### Features
- Ability to link view all pages via `link.viewall-group-all` and `link.viewall-group-subgroup`
#### Bugs
- Fixed test cases incorrect
- Fixed view all pages not generated
- Fixed view all page could not be loaded
- Fixed pattern ordering
- Fixed subfolder patterns have invalid data
- Fixed subfolder patterns are not rendered properly
- Fixed deeper nesting patterns are ignored
- Fixed deeper nesting patterns are not rendered properly


packages/core/patternlab-config.json
```diff
@@ -75,6 +75,10 @@
   "patternExportPatternPartials": [],
   "patternExportPreserveDirectoryStructure": true,
   "patternExportRaw": false,
+  "patternTranslations": {
```
JosefBredereck, line 78: I'm not quite sure if this is a good approach but giving the user the ability to change the automated generated texts seems to be a good idea.

packages/core/src/lib/object_factory.js
```diff
@@ -1,108 +1,87 @@
...
-  // this is the unique name, subDir + fileName (sans extension)
-  this.name = '';
-  if (info.hasDir && info.dirLevel > 2) {
-    let variant = '';
-
-    if (this.fileName.indexOf('~') !== -1) {
-      variant = '-' + this.fileName.split('~')[1];
-    }
-    this.name = this.subdir.replace(/[\/\\]/g, '-') + variant;
+  // TODO: Remove if when droping ordering by prefix and keep else code
```
JosefBredereck: Sometimes when we change the ordering by prefix, we can also reduce some code here.

packages/core/src/lib/object_factory.js
```diff
@@ -125,7 +117,7 @@ const Pattern = function(relPath, data, patternlab) {
   this.lineageR = [];
   this.lineageRIndex = [];
   this.isPseudoPattern = false;
-  this.order = Number.MAX_SAFE_INTEGER;
+  this.order = 0;
```
JosefBredereck, line 131: Patterns should be sorted after groups and before view all page of that group.

packages/core/src/lib/object_factory.js
```diff
@@ -218,40 +210,98 @@ Pattern.prototype = {
...
     if (items[level]) {
       return items[level];
-    } else if (items[level - 1]) {
-      return items[level - 1];
-    } else {
+    } else if (level >= 1) {
       return '';
+    } else {
+      // Im Not quite shure about that but its better than empty node
+      // TODO: verify
```
JosefBredereck, line 249: User can set the root group name in `patternlab-config`

packages/core/test/object_factory_tests.js
```diff
@@ -198,32 +203,79 @@ tap.test(
...
-    test.equals(p.getDirLevel(3, { hasDir: true, dirLevel: 1 }), '');
+    test.equals(p.name, '00-atoms-00-global-00-colors-alt-variant');
+    test.equals(p.flatPatternPath, '00-atoms-00-global');
+    test.equals(p.patternBaseName, 'colors-alt-variant');
+
     test.end();
   }
 );
 
+tap.test('test Patterns that are nested deeper without own directory', function(
```
JosefBredereck, line 266: I added additional test cases here just to be sure if changes are made everything is working correctly after that.

packages/core/test/object_factory_tests.js
```diff
@@ -79,19 +79,21 @@ tap.test(
         path.sep +
         'colors.mustache'
     );
-    test.equals(p.name, '00-atoms-00-global');
```
JosefBredereck, line 82: Reversed previously made changes

packages/core/test/ui_builder_tests.js
```diff
@@ -183,6 +183,7 @@ tap.test('groupPatterns - creates pattern groups correctly', function(test) {
   });
 
   patternlab.patterns.push(
+    new Pattern('foobar.mustache'),
```
JosefBredereck, line 186: Added an additional flat pattern

packages/core/test/ui_builder_tests.js
```diff
@@ -222,13 +223,19 @@ tap.test('groupPatterns - creates pattern groups correctly', function(test) {
     'patternType1-white'
   );
 
+  // Flat patterns
   test.equals(
```
JosefBredereck, line 227: Additional test for flat patterns

packages/core/test/ui_builder_tests.js
```diff
@@ -248,13 +255,15 @@ tap.test('groupPatterns - orders patterns when provided from md', function(
     subtypePatterns: {},
   });
 
+  // Should be sorted by order and secondly by name
   patternlab.patterns.push(
```
JosefBredereck, line 259: Changed the ordering while creation to demonstrate that ordering by order + name works

packages/core/test/ui_builder_tests.js
```diff
@@ -546,40 +555,42 @@ tap.test(
...
+      // when what was intended was a continue
+      // we expect 8 here because:
+      //   - foo.mustache is flat and therefore does not have a viewall page
+      //   - the colors.mustache files make 6
+      //   - patternSubType1 and patternSubType2 make 8
+      // while most of that heavy lifting occurs inside groupPatterns and not buildViewAllPages,
+      // it's important to ensure that this method does not get prematurely terminated
+      // we choose to do that by checking it's return number of patterns
+
+      const uniquePatterns = ui.uniqueAllPatterns(allPatterns, patternlab);
```
JosefBredereck, line 648: Updated to use real code function instead of dummy

packages/core/src/lib/ui_builder.js
```diff
@@ -258,32 +253,16 @@ const ui_builder = function() {
    * Creates a patternSubTypeItem object from a pattern
    * This is a menu item you click on
    * @param pattern - the pattern to derive the subtypeitem from
-   * @returns {{patternPartial: string, patternName: (*|string), patternState: string, patternSrcPath: string, patternPath: string}}
+   * @returns {{patternPartial: string, patternName: (*|string), patternState: string, patternPath: string}}
    */
   function createPatternSubTypeItem(pattern) {
-    let patternPath = '';
```
JosefBredereck, line 264: Patterns generate already a pattern link so that code was not required

packages/core/src/lib/ui_builder.js
```diff
@@ -621,96 +613,92 @@ const ui_builder = function() {
...
-              .then(() => {
-                //do not create a viewall page for flat patterns
-                if (!writeViewAllFile || !p) {
+              .catch(reason => {
+                console.log(reason);
+                logger.error('Error building footer HTML');
+              });
+          }
+        );
+
```
JosefBredereck, line 623: Moved creation of view all pages down here because the files where generated multiple times. With this approach, it will only happen once for every view all page.

packages/core/src/lib/ui_builder.js
```diff
@@ -722,15 +710,41 @@ const ui_builder = function() {
...
   /**
    * Reset any global data we use between builds to guard against double adding things
+   *
+   * @param {Patternlab} patternlab Actual patternlab instance
    */
   function resetUIBuilderState(patternlab) {
     patternlab.patternPaths = {};
     patternlab.viewAllPaths = {};
     patternlab.patternTypes = [];
   }
```
JosefBredereck, line 732: I created a function to also test the functionality and not just fake it ;)

JosefBredereck:
@bmuenzenmeyer I have also added some comments in here [1143/files](https://github.com/pattern-lab/patternlab-node/pull/1143/files) for lines or blocks where it seems to be necessary. Feel free to add additional comments.

packages/core/patternlab-config.json
JosefBredereck: I'm not quite sure if this is a good approach but giving the user the ability to change the automated generated texts seems to be a good idea.

bmuenzenmeyer:
@JosefBredereck excited to look at this closer - but I need a day or so more

JosefBredereck:
Take your time. This should be checked completely and its not just one line of code I changed.

coreylafferty:
@JosefBredereck I gave this a quick test and it does appear to fix #1134.  Thank you!  I'm not sure if it's an intentional change or not, but I did notice that any categories with nested items now appear at the top of the list before those without children.  Just thought I'd draw attention to it in case that's not the intended behavior.

JosefBredereck:
> @JosefBredereck I gave this a quick test and it does appear to fix #1134. Thank you! I'm not sure if it's an intentional change or not, but I did notice that any categories with nested items now appear at the top of the list before those without children. Just thought I'd draw attention to it in case that's not the intended behavior.

I will test that again explicit, thanks for the hint. 👍

JosefBredereck:
@coreylafferty maybe that's what I get and it looks correct to me. When you add a pattern in the same folder as the pattern name the folder will get removed. The folders are for better structuring in your folder architecture. See #992 

grafik

And when I click on broken item 1 it's also correct.

grafik

JosefBredereck:
@coreylafferty that's what I get and it looks correct to me. When you add a pattern in the same folder as the pattern name the folder will get removed. The folders are for better structuring in your folder architecture. See #992 - Also if you add something like `atoms/atoms.twight`, that wouldn't make much sense.

grafik

And when I click on broken item 1 it's also correct.

grafik

coreylafferty:
@JosefBredereck I just tested it again and I am seeing some odd behavior.  Below is a screenshot of our theme with your changes applied.
 
menu

I'd expected to see 'Article' come before 'Button', but instead all of the items with children are grouped at the top of the list followed by single items.  That may be the intention so maybe that doesn't need to change...  but there is a bigger problem.

More troubling is that now 'Button' is both a folder and a single item.  I would expect that the single 'Button' item would be in the 'Button' folder with the variants (e.g., Danger Button, Large Button, etc.).

The way we organize our files is according to BEM class names, so that's why this is important.  For example, here is how Article and Button are structured in our folders:
```
patterns/04-components/article/article.twig
patterns/04-components/button/button.twig
patterns/04-components/button/button--danger/button--danger.twig
patterns/04-components/button/button--large/button--large.twig
```

For reference, here's what our theme looks like using Pattern Lab version 5.4.0 (prior to the change that broke things for us)

menu-5 4

JosefBredereck:
Let me think about that until tomorrow. Maybe I can get a good solution for that. 🤔

coreylafferty:
> Let me think about that until tomorrow. Maybe I can get a good solution for that. 🤔 Very good case by the way. 

Thank you! Please let me know if I can give you any more details or test anything.

JosefBredereck:
I came across some ideas to handle this issue. 

1. I thought the easiest way would be to change the name of the button to buttons, but that wouldn't be a forward-looking solution for the future.
2. I thought that you could put the button file in another button folder, but `/button/button/button.twig` wouldn't be very pretty and also wouldn't be a forward-looking solution for the future.
3. I came across a solution that is already used for another feature. MD-File data on how they are used in [pattern-states](https://patternlab.io/docs/pattern-states.html). So after thinking about in which way I should implement the functionality, there were two options:
- `subbed` with default value `true`
- `subbed` with default value `false`
In my opinion, it would be more common that users (maybe because I'm using only this format) will use the subbed version more often than the unsubbed. So I set the `subbed` value to `true` by `default`.


How to remove the pattern-subbing for your pattern?
in your `button.md` file just insert the following lines at the beginning.
```
---
subbed: false
---
```

And your button will not be handled as a pattern with its own pattern directory.

coreylafferty:
> I came across some ideas to handle this issue.
>     3. I came across a solution that is already used for another feature. MD-File data on how they are used in [pattern-states](https://patternlab.io/docs/pattern-states.html). So after thinking about in which way I should implement the functionality, there were two options:


I just tested the changes for your third solution and the 'subbed: false' feature seems to work exactly as we'd need.  It'd be more convenient for our individual situation to have 'false' be the default (and I think that would be more in line with how Pattern Lab has behaved in the past), but I will admit I don't have a full grasp on the expectations for these new changes, so maybe most users would prefer it be set to 'true'.  

I'd love to hear others thoughts on the impact of these changes.  For the purposes of our theme we just wanted the folder organization to work the same as the original PHP version (which is currently how things work prior to the problem introduced in 5.7, which is why we've currently avoided upgrading past 5.4).  If the same organization as the PHP version / how 5.4 works is not the goal though and these collapsed items are the default desired behavior, then the 'subbed' option in the MD file seems like an easy way to get us to where we need to be.  Thank you for looking into this!

JosefBredereck:
I would like to request some thoughts about that from @bmuenzenmeyer and @sghoweri because I can't decide it by myself. Maybe they have an idea or even a better approach for that. The default value is not set in stone right now.

coreylafferty:
@JosefBredereck i was thinking about this a bit more and had a question... why is the new variable called ‘subbed’ and disabling it puts the item in a subfolder? I’m probably just not understanding the purpose/name fully, but offhand I’d think of naming it ‘subfolder’ and making the true state behave as the current false condition does. 

Curious to hear what @bmuenzenmeyer @sghoweri or others think about all these changes/fixes and the default behavior of the nesting folders. Currently we can’t upgrade the version of PL in our theme without things breaking, but if this gets accepted we should be unblocked. Thank you!

JosefBredereck:
> @JosefBredereck I was thinking about this a bit more and had a question... why is the new variable called ‘subbed’ and disabling it puts the item in a subfolder? I’m probably just not understanding the purpose/name fully, but offhand I’d think of naming it ‘subfolder’ and making the true state behave as the current false condition does.

I would say `subbed` was just the short version of `subfolder`. For the suggestion: Yes I will take that into account. As mentioned before nothing is set in stone and we have to decide together what works best. We have **two votes for** `subfolder = false` as default and one against so far.

sghoweri:
> @JosefBredereck I just tested it again and I am seeing some odd behavior. Below is a screenshot of our theme with your changes applied.
> 
> menu
> 
> I'd expected to see 'Article' come before 'Button', but instead all of the items with children are grouped at the top of the list followed by single items. That may be the intention so maybe that doesn't need to change... but there is a bigger problem.
> 
> More troubling is that now 'Button' is both a folder and a single item. I would expect that the single 'Button' item would be in the 'Button' folder with the variants (e.g., Danger Button, Large Button, etc.).
> 
> The way we organize our files is according to BEM class names, so that's why this is important. For example, here is how Article and Button are structured in our folders:
> 
> ```
> patterns/04-components/article/article.twig
> patterns/04-components/button/button.twig
> patterns/04-components/button/button--danger/button--danger.twig
> patterns/04-components/button/button--large/button--large.twig
> ```
> 
> For reference, here's what our theme looks like using Pattern Lab version 5.4.0 (prior to the change that broke things for us)
> 
> menu-5 4

@JosefBredereck Is this ready for review or are are still some kinks getting worked out?

packages/core/patternlab-config.json
sghoweri: We might be able to come up with a better name for what we're calling these but in general I'm totally onboard with adding support for internationalization / localization.

packages/core/patternlab-config.json
```diff
@@ -75,6 +75,10 @@
...
+    "view-all": "View all {groupName}",
+    "root-name": "Root"
```
sghoweri: Can we make these camelCase vs kebab-case?

JosefBredereck:
> @JosefBredereck Is this ready for review or are still some kinks getting worked out?

@sghoweri yes there is one open discussion at the moment. It's the subfolder's default behavior.
It's subfolder patterns as default or should they be treated as normal patterns and you can turn on the subfolder in markdown.

packages/core/patternlab-config.json
JosefBredereck:
```json
"viewAll": "View all {groupName}",
"rootName": "Root"
```

Like this?

packages/core/patternlab-config.json
JosefBredereck: I just saw that there will be an issue in the pl-nav maybe. There are some cases where checks are done against the name of the current pattern `patternSubtypeItem.patternName === 'View All'`.

packages/core/patternlab-config.json
JosefBredereck: I supplied a change

packages/core/patternlab-config.json
JosefBredereck: I fixed this issue in the last commit

coreylafferty:
@sghoweri Yes, I think this is ready for review.  There's the question of the default behavior for the nav subfolders, but I think @JosefBredereck is looking for feedback from you and @bmuenzenmeyer on that as part of the review (please correct me if I'm wrong, @JosefBredereck).

JosefBredereck:
@sghoweri @bmuenzenmeyer I would kindly ask for a review of these changes, because there is a bigger project starting in 3 weeks where I need the changes made in this pr.

Also I know there is still one open discussion but it is a small one that can also be handled afterwards.

Thanks in advance 🙂

bmuenzenmeyer:
I've started a review of this. Documenting that I think this functionality should be released as a breaking change. Flat patterns (think templates and pages) previously did not display View All Pages by design.

packages/core/src/lib/ui_builder.js
```diff
@@ -621,96 +609,92 @@ const ui_builder = function() {
...
+                return buildViewAllHTML(
                   patternlab,
-                  'viewall-' + patternType + '-all',
+                  typePatterns,
+                  patternType,
                   uikit
                 )
-                  .then(footerHTML => {
-                    //add any flat patterns
-                    //todo this isn't quite working yet
```
JosefBredereck, line 641: I followed this todo for flat patterns. Maybe it was a false assumption.

packages/core/src/lib/object_factory.js
```diff
@@ -1,108 +1,88 @@
...
-const patternPrefixMatcher = /^_?(\d+-)?/;
+const prefixMatcher = /^_?(\d+-)?/;
 
-// Pattern properties
 /**
- * Pattern constructor
- * @constructor
+ * Pattern constructor / Pattern properties
+ *
+ * Before changing functionalities of the pattern object please read the following pull requests
```
bmuenzenmeyer, line 15: this is a smart comment for the next poor soul

packages/core/src/lib/object_factory.js
```diff
@@ -1,108 +1,88 @@
...
+ * to get more details about the behavior of the folder structure
+ * https://github.com/pattern-lab/patternlab-node/pull/992
+ * https://github.com/pattern-lab/patternlab-node/pull/1016
+ * https://github.com/pattern-lab/patternlab-node/pull/1143
+ *
+ * @param {String} relPath relative directory
```
bmuenzenmeyer: usually these jsdoc types should be the literals, like `string` or `boolean`, not the Object `String`

packages/core/src/lib/object_factory.js
```diff
@@ -1,108 +1,88 @@
...
+ * @param {Object} jsonFileData The JSON used to render values in the pattern.
+ * @param {Patternlab} patternlab The actual patternlab instance
+ * @param {Boolean} isUnsubRun specifies if the pattern needs to be removed from its subfolder
```
bmuenzenmeyer: this parameter name is not at all clear to me. it be renamed?

packages/core/src/lib/object_factory.js
```diff
@@ -1,108 +1,88 @@
...
-  // this is the unique name, subDir + fileName (sans extension)
-  this.name = '';
-  if (info.hasDir && info.dirLevel > 2) {
-    let variant = '';
-
-    if (this.fileName.indexOf('~') !== -1) {
-      variant = '-' + this.fileName.split('~')[1];
-    }
-    this.name = this.subdir.replace(/[\/\\]/g, '-') + variant;
+  // TODO: Remove if when droping ordering by prefix and keep else code
```
bmuenzenmeyer: `dropping`

packages/core/src/lib/object_factory.js
```diff
@@ -1,108 +1,88 @@
...
-        ' ' +
-        working.charAt(0).toUpperCase() +
-        working.slice(1)
-      );
-    }, '')
-    .trim(); //this is the display name for the ui. strip numeric + hyphen prefixes
+  // Fancy name - Uppercase letters of pattern name partials.
+  // global-colors -> 'Global Colors'
+  // this is the display name for the ui. strip numeric + hyphen prefixes
+  this.patternName = _.startCase(this.patternBaseName);
```
bmuenzenmeyer, line 76: WOW what an improvement over my terrible code from years ago 😄

packages/core/src/lib/object_factory.js
```diff
@@ -112,11 +92,24 @@ const Pattern = function(relPath, data, patternlab) {
 
   // Let's calculate the verbose name ahead of time! We don't use path.sep here
   // on purpose. This isn't a file name!
-  this.verbosePartial =
-    this.subdir.split(path.sep).join('/') + '/' + this.fileName;
+  this.verbosePartial = `${info.shortNotation}/${this.fileName}`;
+
+  /**
+   * Definition of flat pattern:
```
bmuenzenmeyer, line 108: this is helpful - though i wonder if the entire convention should have been done away with at some point to avoid special casing this throughout

packages/core/src/lib/object_factory.js
```diff
@@ -218,40 +220,123 @@ Pattern.prototype = {
...
 
     if (items[level]) {
       return items[level];
-    } else if (items[level - 1]) {
-      return items[level - 1];
-    } else {
+    } else if (level >= 1) {
       return '';
+    } else {
+      // Im Not quite shure about that but its better than empty node
```
bmuenzenmeyer: I'm not quite sure

packages/core/src/lib/object_factory.js
```diff
@@ -218,40 +220,123 @@ Pattern.prototype = {
...
   },
+
+  /**
+   * Reset the information that the pattern has it's own directory,
+   * so that this pattern will not be handled as flat pattern if it
+   * is located on a top level folder.
+   *
+   * @param {Patternlab} patternlab Current patternlab instance
+   */
+  resetSubbing: function(patternlab) {
```
bmuenzenmeyer: if my working understanding is correct - this could instead be called something like `moveFlatPatternToSubfolder`, yes?

packages/core/src/lib/object_factory.js
```diff
@@ -218,40 +220,123 @@ Pattern.prototype = {
...
+      flatPatternPath: p.flatPatternPath,
+      patternPartial: p.patternPartial,
+      verbosePartial: p.verbosePartial,
+    });
+  },
+
+  /**
+   * Info contains information about pattern structure if it is a nested pattern
+   * or if it just a subfolder structure. Its just used for internal purposes.
+   * Remember every pattern infomarion based on "this.*" will be used by other functions
```
bmuenzenmeyer: information

packages/core/src/lib/object_factory.js
```diff
@@ -218,40 +220,123 @@ Pattern.prototype = {
...
+   *
+   * @param pathObj path.parse() object containing usefull path information
```
bmuenzenmeyer: useful

packages/core/src/lib/readDocumentation.js
```diff
@@ -54,6 +54,14 @@ module.exports = function(pattern, patternlab) {
       if (markdownObject.links) {
         pattern.links = markdownObject.links;
       }
+
+      if (
+        markdownObject.hasOwnProperty('subfolder') &&
+        !markdownObject.subfolder
```
bmuenzenmeyer: i tested this exhaustively with the commented `subbed` key until finally finding this - oops!

JosefBredereck:
> Pardon some of the comments pointing out typos, I stopped after a bit. It's not fair to nitpick when it's 1) possible English is not your first language (I highly respect the idea of programming in a non-native language), and 2) something we could easily automate with a VS Code extension or linter.

I'm sorry. It's true, English is not my native language. I can check the code and comments with Grammarly or something similar if you want to.

> Okay - I've read through all this a couple of times, tested it, and have a better understanding of what's all going on regarding subfolder. Before I get to it - I find it useful to document the steps I took to perform my validations, as the generated handlebars preview doesn't cover it directly.

I try to explain it via screenshots brad frost did in the past and also share a link to the docs. This is really a complicated topic and I hope this makes some things clear.
[Docs page - Deeper Nesting](https://patternlab.io/docs/overview-of-patterns/#heading-deeper-nesting) Most of this is not about buttons.hbs in buttons folder, it's more about restructuring files

### This is what the folder structure is initially intended to look like and I believe most people are working like this.
grafik

### This is what I mostly work with (Except that I don't use numbers for ordering)
grafik


> For what it's worth, I read it as a noun, so the behavior to me seems like the exact opposite of what it should be.
> 
> I expect: (when buttons.hbs exists within buttons/)
> Condition | Outcome
> -- | --
> subfolder not set | Buttons shows up under Buttons Menu
> subfolder: true | Buttons shows up under Buttons Menu
> subfolder: false | Buttons shows up as a "flat pattern"

My intention was that the button is initially handled as a pattern that is deeply nested. And if you don't want this one pattern to be deeply nested you turn it off. Again I was not sure about the naming. Why I used this approach? Because in my opinion, it is more common that you want to structure your patterns with deeper nesting when the names of the folder and pattern match.

> my proposal
> 
>       if (
>         markdownObject.hasOwnProperty('subfolder') &&
>         !markdownObject.subfolder
>       ) {
>         // Reset to pattern without own pattern-directory
>         pattern.promoteFromDirectoryToFlatPattern(patternlab);
>       }

I like this name change.

bmuenzenmeyer:
@JosefBredereck thanks for explaining that to me - even in my example (using the demo), you can see the existing organizational pattern happening, and starting to bloat (especially if one adds all the extra files)

image

As they say, naming is hard! I had always assumed people would make:

atoms/form-input/...
atoms/form-label/...
atoms/form-description-label/...

but it's perfectly natural to also attempt what you are doing too!

Can you tell me, in your screenshot, where do you expect `70-big-radio.hbs` to show up in the nav?

JosefBredereck:
> Can you tell me, in your screenshot, where do you expect 70-big-radio.hbs to show up in the nav?

Would be for all the folders under forms (they are deeply nested patterns)
-> Atoms
--> Forms
----> Input
----> Label
----> Description Label
----> Select
----> Text Area
----> Big Checkbox
----> Big Radio
~ Also variants are a part of that deeply nesting
----> Big Radio Checked Disabled
----> Big Radio Checked
----> Big Radio Disabled

packages/core/src/lib/object_factory.js
JosefBredereck: I would rather go with you other suggestion
`pattern.promoteFromDirectoryToFlatPattern(patternlab);`

coreylafferty:
@bmuenzenmeyer @JosefBredereck Does Josef's answer above point towards next steps and/or are specific changes still requested to this PR?  Hoping to see this accepted (when it's ready) as the fixes here will unblock us from upgrading PL in our theme. For what it's worth, I agree with with @bmuenzenmeyer's assessment of the subfolder variable... I prefer that name (subfolder over subbed) and think when it's true (or not set), items should show in a subfolder.

> Condition | Outcome
> -- | --
> subfolder not set | Buttons shows up under Buttons Menu
> subfolder: true | Buttons shows up under Buttons Menu
> subfolder: false | Buttons shows up as a "flat pattern"

JosefBredereck:
From my point of view there was nothing left so far.

JosefBredereck:
Updated from dev.
I made the flat patterns on view all pages and view all pages for only flat patterns optional. So we don't have a breaking change and if someone wants to turn it on the person is now able to.

JosefBredereck:
> For what it's worth, I read it as a noun, so the behavior to me seems like the exact opposite of what it should be.
> 
> I expect: (when buttons.hbs exist within buttons/)
> Condition | Outcome
> -- | --
> subfolder not set | Buttons shows up under Buttons Menu
> subfolder: true | Buttons shows up under Buttons Menu
> subfolder: false | Buttons shows up as a "flat pattern"

I changed this to a more specific name `deeplyNested` and also updated the documentation regarding this change. So everything will be aligned with [Pattern Overview Deeper Nesting](https://patternlab.io/docs/overview-of-patterns/#heading-deeper-nesting). And we won't have a breaking change again. So we should be aligned with all the discussions going on in this thread.

- **deeplyNested not set or false**  - Pattern won't be handled as a deeply nested pattern
- **deeplyNested: true**  - Pattern will be handled like mentioned under [Deeper Nesting](#heading-deeper-nesting)

sghoweri:
> > For what it's worth, I read it as a noun, so the behavior to me seems like the exact opposite of what it should be.
> > I expect: (when buttons.hbs exist within buttons/)
> > 
> > 
> > 
> > Condition
> > Outcome
> > 
> > 
> > 
> > 
> > subfolder not set
> > Buttons shows up under Buttons Menu
> > 
> > 
> > subfolder: true
> > Buttons shows up under Buttons Menu
> > 
> > 
> > subfolder: false
> > Buttons shows up as a "flat pattern"
> 
> I changed this to a more specific name `deeplyNested` and also updated the documentation regarding this change. So everything will be aligned with [Pattern Overview Deeper Nesting](https://patternlab.io/docs/overview-of-patterns/#heading-deeper-nesting). And we won't have a breaking change again. So we should be aligned with all the discussions going on in this thread.
> 
> * **deeplyNested not set or false**  - Pattern won't be handled as a deeply nested pattern
> * **deeplyNested: true**  - Pattern will be handled like mentioned under [Deeper Nesting](#heading-deeper-nesting)

Hey @JosefBredereck @coreylafferty so sorry for being a bit out of the loop on this -- this good to get merged down and released, yes?

JosefBredereck:
From my point of view, yes. I also used this branch in a production application already. 

I hope the docs are enough and grammatically correct.

coreylafferty:
@sghoweri @JosefBredereck I won't pretend to understand all the changes here, but in my testing this solves the nesting problem I'd described in #1134 and the nesting behaves as I'd expect.  Setting the DeeplyNested value in the MD files does move the pattern in or out of a subfolder, so that's working as described, with the default, undefined condition being how PL has traditionally handled the nesting.  

I'll admit I don't fully understand  what the name 'DeeplyNested' is meant to indicate since at first I expected it to behave the opposite than it does -- with the 'false' condition being the one that flattens the item in the menu structure, placing it outside its own folder (which currently is the True condition)...  but since the default/false is to have the menu behave as I'd expect, I don't see the name being a big issue.

sghoweri:
PR was released with v5.12.0

PR #1148, opened 2020-04-21
fix: actually exit build when Twig render fails

EvanLovely:
Before these changes, when a Twig syntax error was encountered (like using an incorrect function or filter), the error message would be displayed in the Terminal output but would say the build happened ok. The error messages from the Twig Renderer come out like "Error trying to render "01-components/card/05-card.twig". Unknown "render" filter.". After these changes, it displays the error message and then exits non-zero, which stops the output right where the dev can see the issue (either in watch mode or in CI builds).

### DIFF
--- a/packages/engine-twig-php/lib/engine_twig_php.js
+++ b/packages/engine-twig-php/lib/engine_twig_php.js
@@ -91,8 +91,16 @@ const engine_twig_php = {
           if (results.ok) {
             resolve(results.html + details);
           } else {
-            console.log(results.message);
-            process.exit(1);
+            // make Twig rendering errors more noticeable + exit when not in dev mode (or running the `patternlab serve` command)
+            if (
+              process.argv.slice(1).includes('serve') ||
+              process.env.NODE_ENV === 'development'
+            ) {
+              reject(chalk.red(results.message));
+            } else {
+              console.log(chalk.red(results.message));
+              process.exit(1);
+            }
           }
         })
         .catch(error => {
--- a/packages/engine-twig-php/lib/engine_twig_php.js
+++ b/packages/engine-twig-php/lib/engine_twig_php.js
@@ -17,6 +17,7 @@
 const TwigRenderer = require('@basalt/twig-renderer');
 const fs = require('fs-extra');
 const path = require('path');
+const chalk = require('chalk');
 
 let twigRenderer;
 let patternLabConfig = {};
--- a/packages/engine-twig-php/lib/engine_twig_php.js
+++ b/packages/engine-twig-php/lib/engine_twig_php.js
@@ -90,7 +90,8 @@ const engine_twig_php = {
           if (results.ok) {
             resolve(results.html + details);
           } else {
-            reject(results.message);
+            console.log(results.message);
+            process.exit(1);
           }
         })
         .catch(error => {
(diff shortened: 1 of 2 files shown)


EvanLovely:
I usually use an external tool like `nodemon` to watch files and trigger a `patternlab build`, so I'd want that command to exit non-zero. However I did just try `patternlab serve` and that exited before creating a browsable url. 

Ideally an engine can throw an error in the `renderPattern` function it exports and PL core would show that error then either: 1) if `build` exit non-zero, 2) if `serve` keep going. However, that's what I was doing earlier (i.e. how it currently works): is firing the `reject()` of the Promise. So that's why I stepped it up to an exit.

sghoweri:
> I usually use an external tool like `nodemon` to watch files and trigger a `patternlab build`, so I'd want that command to exit non-zero. However I did just try `patternlab serve` and that exited before creating a browsable url.
> 
> Ideally an engine can throw an error in the `renderPattern` function it exports and PL core would show that error then either: 1) if `build` exit non-zero, 2) if `serve` keep going. However, that's what I was doing earlier (i.e. how it currently works): is firing the `reject()` of the Promise. So that's why I stepped it up to an exit.

Totally understood (and I don't disagree with exiting with a non-zero exit code while building for production and/or not directly calling the `patternlab serve` command).

I just want to make sure that folks who are aren't using an external tool like `nodemon` (ie. using the built-in PL CLI) aren't going to have a frustrating experience while building stuff locally...

I've got a proposed tweak to what you have here @EvanLovely that should still get you what you're looking for here in this PR but also retains the ability to work on Twig templates with errors (and not have to continuously start / stop the CLI in the process)!

packages/engine-twig-php/lib/engine_twig_php.js
```diff
@@ -90,7 +90,8 @@ const engine_twig_php = {
           if (results.ok) {
             resolve(results.html + details);
           } else {
-            reject(results.message);
+            console.log(results.message);
+            process.exit(1);
```
sghoweri, line 94:
```suggestion
            // make Twig rendering errors more noticeable + exit when not in dev mode (or running the `patternlab serve` command)
            if (
              process.argv.slice(1).includes('serve') ||
              process.env.NODE_ENV === 'development'
            ) {
              reject(chalk.red(results.message));
            } else {
              console.log(chalk.red(results.message));
              process.exit(1);
            }
```

sghoweri: LMK what you think -- if you're good with this then I'm good with this 👍

sghoweri: `patternlab serve`
image

vs

`patternlab build`
image

bmuenzenmeyer: this looks like a stellar UX to me

bmuenzenmeyer: @EvanLovely i'll let you drive the acceptance of this

EvanLovely: I'm down 👍 - just committed to this branch. I would suggest similar behavior if any template engine called `reject()` from it's `renderPattern()` function. However I think this works great for now. Thanks!

sghoweri: 👍🏼

PR #1159, opened 2020-04-25
chore(docs): some typos

mfranzke:
Summary of changes:
missed to replace mustache initially and incorrect `handlebars` two times (both in package name and URL)

### DIFF
--- a/packages/starterkit-handlebars-demo/README.md
+++ b/packages/starterkit-handlebars-demo/README.md
@@ -4,9 +4,9 @@ The Demo StarterKit for Handlebars is meant to be used as a demonstration of a H
 
 ## Requirements
 
-The Base StarterKit for Mustache requires the following PatternEngine:
+The Vanilla StarterKit for Handlebars requires the following PatternEngine:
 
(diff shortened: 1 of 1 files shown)


packages/starterkit-handlebars-demo/README.md
```diff
@@ -4,9 +4,9 @@ The Demo StarterKit for Handlebars is meant to be used as a demonstration of a H
 
 ## Requirements
 
-The Base StarterKit for Mustache requires the following PatternEngine:
+The Vanilla StarterKit for Handlebars requires the following PatternEngine:
```
bmuenzenmeyer, line 7: Is there a reason you changed Base to Vanilla? I think base is probably a better definition for those not familiar with vanilla being "bland" or "normal"

mfranzke: > Is there a reason you changed Base to Vanilla? I think base is probably a better definition for those not familiar with vanilla being "bland" or "normal"

I've adapted it from this file: https://github.com/pattern-lab/starterkit-handlebars-vanilla/blob/master/README.md - but that's incorrect for this part, ob

PR #1165, opened 2020-04-28
Fix plugin tabs enabling multiple file formats

### LINKED ISSUE #1163: fix(plugintabs): enabling multiple file formats
@bmuenzenmeyer your overwrite with commit b82bd129fdbe48de95b62d75fb7fe95cea896b7e of the [placeholders `<<type>>` and `<<typeUC>>` (to hardcoded `scss`)](https://github.com/pattern-lab/patternlab-node/blob/73eac976461f4e587b0c30668942c4895aea319f/packages/plugin-tab/src/snippet.js#L3-L11) that are used within the script https://github.com/pattern-lab/patternlab-node/blob/dev/packages/plugin-tab/index.js now results in that only one file (and only `scss` filetype/language) could get configured within the `patternlab-config.json`.

It looks to me like that this happened by accident - I would than like to offer to fix this (and some other small parts) for this to work correctly again, if the change didn't happen on purpose.

mfranzke:
Closes #1163
Blocked by #1155 #1156 (#1156 has been already merged into this branch, as it's related)

Summary of changes:
Reverted parts to the previously used placeholders within the `/packages/plugin-tab/src/snippet.js` file, and modified `/packages/plugin-tab/dist/js/@pattern-lab-plugin-tab.js` accordingly by migrating some code to that file.

This would reenable defining one or more filetypes to display within the code panels again.


mfranzke:
there seems to have been some problems with the code ... will open another PR

PR #1175, opened 2020-04-30
fix(uikitworkshop): preventing cropping pattern parts #1174

### LINKED ISSUE #1174: fix(uikitworkshop): absolutely positioned pattern parts at the vertical end of the "page" would get cropped
Due to the following CSS declaration, absolutely positioned pattern parts at the end of the "page" would get cropped:
https://github.com/pattern-lab/patternlab-node/blob/123427a2bd0c6285ceddbaaa83c2c90a840ecc54/packages/uikit-workshop/src/sass/pattern-lab.scss#L109

The scroll position doesn't need to be at the end of the viewport; it's mainly related to that the container additionally wouldn't extend to the viewports height, but as well to the initial aspect of cropping absolutely positioned elements on the vertical.

Compare to:
image

Expected display:
image

I've been able to rebuild a simplified example here: https://output.jsbin.com/munabac/1
Only the first item within the "collapsing menu" (after pressing the interaction element "Toggle Overflow menu") is visible within the second pattern.

mfranzke:
absolutely positioned pattern parts at the vertical end of the "page" would get cropped elsewhere

Closes #1174

Summary of changes:
Refactoring parts of commit 123427a as this would elsewhere crop vertically as well, whereas the declaration out of commit 123427a was most likely mainly meant to function horizontally.

Problem:
I've been able to rebuild a simplified example here: https://output.jsbin.com/munabac/1
Only the first item within the "collapsing menu" (after pressing the interaction element "Toggle Overflow menu") is visible within the second pattern.


bmuenzenmeyer:
this seems reasonable - @sghoweri do you have time to review ?

sghoweri:
> this seems reasonable - @sghoweri do you have time to review ?

Sorry, just go around to looking at this. Looks good to me!

PR #1196, opened 2020-05-02
chore(links): prevent redirects both on internal as well as external URLs #1195

### LINKED ISSUE #1195: chore: prevent redirects (and possible MITM attacks)
We could prevent redirects to pages my linking to their "correct" / canonical URL directly.

Additionally not all of the websites we're linking to have introduced the "HTTP Strict Transport Security" (HSTS) mechanism to redirect securely to their HTTPS equivalent. 
Nevertheless we could anyhow prevent that redirect independently from security aspects by linking to their HTTPS version of the website.

Maybe I`m a little too german on this, but at least I've prepared everything for an "easy" merge.

mfranzke:
Some links are missing a trailing slash, and that for causing a redirect. Adding the trailing slash would speed up browsing.
Additionally linking to the correct (HTTPS version) of the pages would provide an additional aspect of "security".

Closes #1195

Summary of changes:
source (partly): https://validator.w3.org/checklink?uri=https%3A%2F%2Fpatternlab.io%2F&hide_type=all&recursive=on&depth=&check=Check

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 This monorepo contains the core of Pattern Lab / Node and all related engines, UI kits, plugins and utilities. Pattern Lab helps you and your team build thoughtful, pattern-driven user interfaces using atomic design principles.
 
-If you'd like to see what a front-end project built with Pattern Lab looks like, check out this [online demo of Pattern Lab output](http://demo.patternlab.io/).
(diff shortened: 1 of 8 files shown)


bmuenzenmeyer:
Thank you so much @mfranzke !!!!

mfranzke:
> Thank you so much @mfranzke !!!!

@bmuenzenmeyer at your service ;-)

PR #1199, opened 2020-05-04
Add possibility for variants merge strategy change

### LINKED ISSUE #1127: Semantic and Data Structure Problems with lodash's merge in Pseudopattern Hunter
I asked this question a week ago in Gitter, but no response so far, so instead of asking this repeatedly in a rythmic fashion, I'll just put this in a ticket:

Hi there,

migrating from Pattern Lab 2 to PL3 I was suprised to find the way pseudo pattern data is constructed being changed. It seems that data is now merged via lodash's _.merge in the pseudo pattern hunter whereas before it was replaced key by key.

I definitely see the appeal of this change for many cases, however this makes practical construction of pattern examples sometimes a little bit difficult and counter-intuitive.

An example:
I have a tabs pattern with 5 tabs as the main json.
Now I want to construct a reduced pseudo variant only showing 2 tabs.

In PL2 I could just overwrite the tabs-array with a new array in the variant, but the PL3 merge just replaces indices one and two and keeps the rest like it is. Of course I could switch the variants, but in my opinion the main json should always show the "default" usage, therefore I am not really free to do this.

Do you have any suggestion on how to handle this?

Of course I could null out single entries in an array, but I'd have to handle that explicitly in template, which again, is cumbersome. The current fact that we can not reset/overwrite a complete array results in technical contraints to logical structures, which should not be the case in my opinion.

In my opinion, the old way of doing a deep extend/assign made total sense, but I also see the pros of the current implementation in certain cases.

My proposal would be this: make the algo configurable via config and maybe allow a pattern-based override.

JosefBredereck:
While variants are now merged with _.merge from lodash there is a need to overwrite this feature. See the closing ticket for more information.

Closes #1127

Summary of changes:
- Add new config param
- Add merge strategy switch


bmuenzenmeyer:
@OktarinTentakel have you had an opportunity to test this? I value your test output on a larger pattern lab

bmuenzenmeyer (changes requested):
some todos before merge - @JosefBredereck isn't necessarily on the hook to do them, I just want to capture them

- [ ] the current behavior should either be the default (`patternMergeVariantArrays`: `false`). or this needs to be marked a breaking change
- [ ] needs tests
- [ ] needs documentation within https://patternlab.io/docs/editing-the-configuration-options/

OktarinTentakel:
@bmuenzenmeyer 

Wouldn't the current default be "patternMergeVariantArrays: true"? Because _.merge _does_ merge arrays, while the old solution _does not_ (the old solution replaced the array at the key).

JosefBredereck:
> - [ ] the current behavior should either be the default (`patternMergeVariantArrays`: `false`). or this needs to be marked a breaking change

The current behavior is that the arrays in variants get merged. If you mean that the parameter maybe is not available then we need to change the naming and also reverse the default behavior.

bmuenzenmeyer:
my bad 😊 - it was exactly 10 minutes ago I read about `mergeWith` for the first time. I'll trust both of you - I was merely trying to point out that we shouldn't ping pong between one strategy or another if there is an existing pattern on the current 5.x core release line.

packages/core/src/lib/pseudopattern_hunter.js
```diff
@@ -60,10 +60,18 @@ pseudopattern_hunter.prototype.find_pseudopatterns = function(
...
-      variantFileData = _.merge(
+      variantFileData = _.mergeWith(
         {},
         currentPattern.jsonFileData,
-        variantFileData
+        variantFileData,
+        (objValue, srcValue) => {
+          if (
+            _.isArray(objValue) &&
+            patternlab.config.patternMergeVariantArrays
```
OktarinTentakel: If this is supposed the replicate the old behaviour, this needs to be !patternlab.config.patternMergeVariantArrays, since the old solution does _not_ merge arrays, but replaces the original array with the variant array.

PS: this also means flipping the config default value

packages/core/src/lib/pseudopattern_hunter.js
```diff
@@ -60,10 +60,18 @@ pseudopattern_hunter.prototype.find_pseudopatterns = function(
...
+          ) {
+            return objValue.concat(srcValue);
```
OktarinTentakel: If I'm not completely misstaken the PL2-way of doing this wasn't to concat but to replace at the key, so i guess this should be "return srcValue;" to only take the variant data and drop the parent array.

Otherwise we'd be again in a situation where we cannot reduce a default pattern's array contents since this time it wouldn't be replaced index by index, but simply extended each and every time :D

JosefBredereck: Now I'm confused by myself. I need to test again in the evening to give an answer about what is happening here. I also should add a comment on that lines for other devs in the future.

OktarinTentakel:
```
const mergeData = function (obj1, obj2) {
  /*eslint-disable no-param-reassign, guard-for-in*/
  if (typeof obj2 === 'undefined') {
    obj2 = {};
  }
  for (var p in obj1) {
    try {
      // Only recurse if obj1[p] is an object.
      if (obj1[p].constructor === Object) {
        // Requires 2 objects as params; create obj2[p] if undefined.
        if (typeof obj2[p] === 'undefined') {
          obj2[p] = {};
        }
        obj2[p] = mergeData(obj1[p], obj2[p]);

        // Pop when recursion meets a non-object. If obj1[p] is a non-object,
        // only copy to undefined obj2[p]. This way, obj2 maintains priority.
      } else if (typeof obj2[p] === 'undefined') {
        obj2[p] = obj1[p];
      }
    } catch (e) {
      // Property in destination object not set; create it and set its value.
      if (typeof obj2[p] === 'undefined') {
        obj2[p] = obj1[p];
      }
    }
  }
  return obj2;
};
```

This is the old merge method (from version 2.12). It just recurses the parent object and sets every nested key if not already present in variant. So variant data always has priority and will never be overriden/merged/concatenated by the parent.

Example:

let parent = {a : 1, b : [2 , 3], c : {d : [4, 5], e : 8}};
let variant = {a : 2, c : {d : [6, 7]}};
mergeData(parent, variant);
=> {a : 2, b : [2 , 3], c : {d : [6, 7], e : 8}};

JosefBredereck: That is what I'm also known to from 2.x. But currently, it is exactly the opposite of the current version. 

So the default should be that variants arrays merge and that you can turn it off.

Now that I tested the mechanism again, it is correct that the functionality was not correct. It should be like the following code. (I pushed an update)

```javascript
if (
  _.isArray(objValue) &&
  !patternlab.config.patternMergeVariantArrays
) {
  return srcValue;
}
```

JosefBredereck:
Tests added documentation follows.

JosefBredereck:
Documentation added, please do a grammar- and spell-check on the documentation too @bmuenzenmeyer

bmuenzenmeyer:
@JosefBredereck @OktarinTentakel great progress on this - if I am reading this correctly - this is a breaking change of 5.X functionality, correct? I am fine with that if it's a restoration of previous functionality, but from a pure semver standpoint - 5.X pseudo pattern data is now merging differently by default, yes?

packages/docs/src/docs/pattern-pseudo-patterns.md
```diff
@@ -19,6 +19,12 @@ Pseudo-patterns are similar to [pattern-specific JSON files](/docs/creating-patt
 
 The tilde (`~`) and `.json` file extension are the hints that Pattern Lab uses to determine that this is a pseudo-pattern. The `patternName` tells Pattern Lab which existing pattern it should use when rendering the pseudo-pattern. The JSON file itself works exactly like the [pattern-specific JSON file](/docs/creating-pattern-specific-values/). It has the added benefit that the pseudo-pattern will also inherit any values from the existing pattern's pattern-specific JSON file.
 
+As default arrays in pseudo-pattern data will be concatenated with the base patterns array data. To override this behavior globally you can set the parameter `patternMergeVariantArrays` in you `patternlab-config.json` to `false`. Arrays will then be overwritten and you will only have the variants array data left.
```
bmuenzenmeyer:
```suggestion
By default, arrays in pseudo-pattern data will be concatenated with the base pattern's array data. To override this behavior globally you can set the parameter `patternMergeVariantArrays` in your `patternlab-config.json` to `false`. Arrays will then be overwritten and you will only have the variant's array data left.
```

bmuenzenmeyer (changes requested):
Thank you for adding tests!

An alphabetical entry also needs to be added to ⁠⁠https://github.com/pattern-lab/patternlab-node/blob/dev/packages/docs/src/docs/advanced-config-options.md

OktarinTentakel:
@bmuenzenmeyer 

As long as patternMergeVariantArrays is true by default, this should not be breaking, since merging arrays is the new way of PL3, while not merging them (but replacing them) ist the old.

So, if this is true by default we should be fine.

OktarinTentakel:
@bmuenzenmeyer 

"By default, arrays in pseudo-pattern data will be concatenated with the base pattern's array data."

No, I don't think they will. _.merge in default mode merges by index.

An example:

let parent = [1, 2, 3, 4, 5];
let variant = [6, 7];
_.merge(parent, variant);
=> [6, 7, 3, 4, 5]

OktarinTentakel:
Just tested this in the lodash documentation, behaves as expected:

merge-test

So the sentence should read:

"By default, arrays in pseudo-pattern data will be merged with the base pattern's array data on an index basis, effectively creating an array having as many entries as the larger of the two, with the indices declared in the variant replacing the first n of the parent, while keeping the rest. Declaring a larger array in the variant results in only using the variant array."

JosefBredereck:
> @bmuenzenmeyer 
> 
> As long as patternMergeVariantArrays is true by default, this should not be breaking, since merging arrays is the new way of PL3, while not merging them (but replacing them) ist the old.
> 
> So, if this is true by default we should be fine.

I also added the case, that if the config param is not available then it should also use the bahavior of patternMergeVariantArrays = true

JosefBredereck:
> @bmuenzenmeyer 
> 
> "By default, arrays in pseudo-pattern data will be concatenated with the base pattern's array data."
> 
> No, I don't think they will. _.merg in default mode merges by index.
> 
> An example:
> 
> let parent = [1, 2, 3, 4, 5];
> let variant = [6, 7];
> _.merge(parent, variant);
> => [6, 7, 3, 4, 5]


Thaks for that hint. In that case I have to adjust the tests and the docs. As I was testing I thought the default `mergeWith` behaves differently to `merge` and added a concat function. Now where I got confirmation that this was really the behavior I can remove the concat.

OktarinTentakel:
@JosefBredereck 

Oh, I'm sorry, I didn't see the second if statement in the pattern hunter. My bad. In that case, the documentation is completely right and @bmuenzenmeyer is right as well. If you concat arrays in the with function by default, this is indeed a breaking change, since this is new and before you used the merge default behaviour I outlined above.

So, if you want to keep the concat behaviour you definitely have to keep the concat in the with function, otherwise you'll have merging again!

Sorry for the confusion, reading helps sometimes. -.-

JosefBredereck:
> Oh, I'm sorry, I didn't see the second if statement in the pattern hunter. My bad. In that case, the documentation is completely right and @bmuenzenmeyer is right as well. If you concat arrays in the with function by default, this is indeed a breaking change, since this is new and before you used the merge default behavior I outlined above.
> 
> So, if you want to keep the concat behavior you definitely have to keep the concat in the with function, otherwise you'll have merging again!

No need to apologize. I wasn't 100% sure about the merge functionality and you helped by testing t again. And also we don't want to make breaking changes today the second statement shouldn't be implemented.

JosefBredereck:
Docs added and adjusted

clairesunstudio:
@JosefBredereck @OktarinTentakel when will this get merged? we are waiting on it to do the upgrade. thanks!

JosefBredereck:
@clairesunstudio We are not allowed to merge into the dev branch and make a new release.

 @bmuenzenmeyer and @sghoweri could do the trick 😁

clairesunstudio:
@bmuenzenmeyer and @sghoweri any plans for releasing this change?

sghoweri:
> @bmuenzenmeyer and @sghoweri any plans for releasing this change?

@clairesunstudio so sorry for being AWOL -- things have been super busy on my end (just started a new job a couple weeks ago). Will try to get this released tomorrow!

mfranzke:
@sghoweri congratulations for your new job and all the best!

sghoweri (approved):
HUGE +1 on including tests _and_ docs with this update! 🎉 💯

sghoweri:
PR was released with v5.11.1

clairesunstudio:
@sghoweri congrats on your new job!! and thanks for the release!

PR #1202, opened 2020-05-09
Fix broken "open in new tab" link in UIKit 

sghoweri:
Just noticed this was broken while reviewing a separate PR... this PR should fix the "Open In New Tab" link in Pattern Lab's context menu so it now opens the correct HTML file vs a broken "undefined" page. 

**Before (Broken)**
CleanShot 2020-05-09 at 14 20 23

**After (Working)**
CleanShot 2020-05-09 at 14 21 01

### DIFF
--- a/packages/uikit-workshop/src/scripts/lit-components/pl-viewport/pl-viewport.js
+++ b/packages/uikit-workshop/src/scripts/lit-components/pl-viewport/pl-viewport.js
@@ -523,6 +523,10 @@ class IFrame extends BaseLitComponent {
           addressReplacement
         );
 
+        const currentUrl = urlHandler.getFileName(currentPattern);
+        if (currentUrl) {
+          store.dispatch(updateCurrentUrl(currentUrl));
+        }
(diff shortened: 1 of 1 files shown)


sghoweri:
@bmuenzenmeyer or @JosefBredereck any chance one of you guys could do a super quick review of this UIKit hotfix?

JosefBredereck:
Yes, it is, I already implemented it in #1143 so I know it will work. Also, you added in addition an check if the URL is available, which I really like. I will remove it from #1143 because it is some lines above this one.

JosefBredereck (approved):

sghoweri:
> Yes, it is, I already implemented it in #1143 so I know it will work. Also, you added in addition an check if the URL is available, which I really like. I will remove it from #1143 because it is some lines above this one.

Ah, I totally missed that this fix was already in #1143 (good catch!). Thanks @JosefBredereck!

sghoweri:
Released in v5.10.1

PR #1211, opened 2020-05-15
fix: enable partial build via option

### LINKED ISSUE #1210: Partial build option in Node doesn't work
I am using @pattern-lab/core `v5.10.1` on `Mac`, with Node `v12.16.1`, not using an Edition.

##### Expected Behavior
Setting the option `cleanPublic` to false via the build function should do a partial build.

##### Actual Behavior
The option is not used, so the function always runs a full build.

##### Steps to Reproduce
In a node or gulp script:

```
const pl = patternlabNode(config);
pl.build({
  cleanPublic: false,
});
```

portfolioris:
Closes #1210

Summary of changes:
Add option to consume partial build option.


sghoweri (approved):
We'll want to update the UIKit internal Webpack build + the internal Handlebars demo (both used for testing) to see this update work, however we can get that tiny update in after this is merged down.

Thanks for the fix @portfolioris!

sghoweri:
PR was released with v5.11.1

PR #1225, opened 2020-06-16
Better support for Yarn workspaces

### LINKED ISSUE #1136: Allow uikit plugins to be used across multiple uikit definitions

ringods:
Closes: #1136 

The code has too many places where locations are calculated into a package inside the `node_modules` folder. The assumption is made here that the the `node_modules` folder is in the current working directory (`process.cwd()`). But with Yarn workspaces, this is not always the case: the `node_modules` can be one or two levels higher in the folder structure.

Summary of changes:
* Replace hardcoded path calculations with a function using `require.resolve`.

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -205,7 +205,7 @@ const patternlab_module = function(config) {
     },
 
     /**
-     * Installs plugin already available via `node_modules/`
+     * Installs plugin already available as a package dependency
      *
      * @memberof patternlab
      * @name installplugin
@@ -215,7 +215,6 @@ const patternlab_module = function(config) {
      */
     installplugin: function(pluginName) {
       //get the config
-      const configPath = path.resolve(process.cwd(), 'patternlab-config.json');
       const plugin_manager = new pm();
 
       plugin_manager.install_plugin(pluginName);
@@ -234,7 +233,7 @@ const patternlab_module = function(config) {
     },
 
     /**
-     * Loads starterkit already available via `node_modules/`
+     * Loads starterkit already available as a package dependency
      *
      * @memberof patternlab
      * @name loadstarterkit
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -214,7 +214,6 @@ const patternlab_module = function(config) {
      * @returns {void}
      */
     installplugin: function(pluginName) {
-      //get the config
       const plugin_manager = new pm();
 
       plugin_manager.install_plugin(pluginName);
--- a/packages/core/src/lib/loaduikits.js
+++ b/packages/core/src/lib/loaduikits.js
@@ -30,40 +30,54 @@ module.exports = patternlab => {
   const uikitConfigs = _.filter(patternlab.config.uikits, 'enabled'); // [1]
   uikitConfigs.forEach(uikitConfig => {
     let uikitLocation = null;
-    try {
-      resolvePackageFolder(uikitConfig.name); // [2]
-    } catch (ex) {
-      // Ignore
-    }
-    if (!uikitLocation) {
+    if ('package' in uikitConfig) {
       try {
-        uikitLocation = resolvePackageFolder(
-          `@pattern-lab/${uikitConfig.name}`
-        );
+        uikitLocation = resolvePackageFolder(uikitConfig.package);
       } catch (ex) {
-        // Ignore
-      }
-    }
-    if (!uikitLocation) {
-      try {
-        uikitLocation = resolvePackageFolder(
-          `@pattern-lab/uikit-${uikitConfig.name}`
+        logger.warning(
+          `Could not find uikit with package name ${uikitConfig.package}. Did you add it to the 'dependencies' section in your 'package.json' file?`
         );
-      } catch (ex) {
-        // Ignore
+        return;
       }
+    } else {
+      // For backwards compatibility, name to package calculation is:
+      // 1. name -> name
+      // 2. name -> uikit-name
+      // 3. name -> @pattern-lab/name
+      // 4. name -> @pattern-lab/uikit-name
+      for (const packageName of [
+        uikitConfig.name,
+        `uikit-${uikitConfig.name}`,
+        `@pattern-lab/${uikitConfig.name}`,
+        `@pattern-lab/uikit-${uikitConfig.name}`,
+      ]) {
+        try {
+          uikitLocation = resolvePackageFolder(packageName); // [2]
+        } catch (ex) {
+          // Ignore
+        }
+        if (uikitLocation != null) {
+          uikitConfig.package = packageName;
+          logger.info(`Found uikit package ${packageName}`);
+          break;
+        }
+      }
+      if (uikitLocation == null) {
+        logger.warning(
+          `Could not find uikit with package name ${uikitConfig.name}, uikit-${uikitConfig.name}, @pattern-lab/${uikitConfig.name} or @pattern-lab/uikit-${uikitConfig.name} defined within patternlab-config.json in the package dependencies.`
+        );
+        return;
+      } else {
+        logger.warning(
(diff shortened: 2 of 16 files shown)


ringods:
@sghoweri @bmuenzenmeyer `loaduikits` iterates over the `node_modules` folder for folder names containing `uikit-`. If there are 3 found, they are checked against the config to verify if the uikit is configured and enabled.

https://github.com/pattern-lab/patternlab-node/blob/831c8b0787f41bd842d1a38c83c4c4756fb1da48/packages/core/src/lib/loaduikits.js#L48

https://github.com/pattern-lab/patternlab-node/blob/831c8b0787f41bd842d1a38c83c4c4756fb1da48/packages/core/src/lib/loaduikits.js#L55-L60

In Yarn v2, you can only load dependencies which are explicitly in your `package.json` file. So I'm thinking of updating the code and doing it the other way around: reading the enabled uikits and then `require`-ing them directly. I would error if a uikit is in the PL config, but not defined as a dependency. 

If I do it this way, am I covering the same behaviour then?

bmuenzenmeyer:
> In Yarn v2, you can only load dependencies which are explicitly in your `package.json` file. So I'm thinking of updating the code and doing it the other way around: reading the enabled uikits and then `require`-ing them directly. I would error if a uikit is in the PL config, but not defined as a dependency.
> 
> If I do it this way, am I covering the same behaviour then?

yes all future work should read the config explictly and look for those only - the `node_module` scanning has lead to false-positives like `uikit-polyfills`

ringods:
@bmuenzenmeyer I moved my resolver code from the `cli` package to `core`. After that, the `require` of this code in `cli` fails:

https://github.com/pattern-lab/patternlab-node/pull/1225/checks?check_run_id=868619630#step:4:48

I don't understand what I'm missing. Any idea?

ringods:
@bmuenzenmeyer in commit 04df565, you can see the rewritten version to load the uikits. Without too much effort, the `packagename` can be added as a uikit config property to cover the requirements covered in #1137.

ringods:
`package` property is added in ceccfd8814b638558b5f53446eb6625162a29c49, hereby copying the functionality offered in #1137

packages/core/src/lib/loaduikits.js
```diff
@@ -5,91 +5,110 @@ const _ = require('lodash');
...
+    let uikitLocation = null;
+    if ('package' in uikitConfig) {
+      try {
+        uikitLocation = resolvePackageFolder(uikitConfig.package);
+      } catch (ex) {
+        logger.warning(
+          `Could not find uikit with package name ${uikitConfig.package}. Did you add it to the 'dependencies' section in your 'package.json' file?`
+        );
+        return;
+      }
```
JosefBredereck, line 41: Please provide an example `patternlab-config.json` for that or change the given ones. The user shouldn't get warnings for a freshly installed pattern-lab after we merge this.

Most Important

- [x] packages/core/patternlab-config.json (used within getDefaultConfig)
- [x] packages/edition-node/patternlab-config.json
- [x] packages/edition-node-gulp/patternlab-config.json
- [x] packages/edition-twig/patternlab-config.json

Less Important or Internal

- [ ] packages/core/test/util/patternlab-config.json
- [x] packages/cli/test/fixtures/patternlab-config.json
- [x] packages/development-edition-engine-handlebars/patternlab-config.json
- [ ] packages/development-edition-engine-react/patternlab-config.json
- [x] packages/development-edition-engine-twig/patternlab-config.json

packages/core/src/lib/resolver.js
```diff
@@ -0,0 +1,33 @@
...
+  return path.dirname(resolveFileInPackage(packageName, ...pathElements));
+};
+
+/**
+ * @func resolvePackageFolder
+ * Resolves the location of a package on disc
+ */
+const resolvePackageFolder = packageName => {
+  return path.dirname(resolveFileInPackage(packageName, 'package.json'));
+};
```
JosefBredereck, line 27: Just a question for this. The whole concept is working without process.cwd() is that correct?
When it tries to resolve the given files/directories it will walk from the resolver location through the filesystem or where would it start?

packages/core/src/index.js
```diff
@@ -215,7 +215,6 @@ const patternlab_module = function(config) {
      */
     installplugin: function(pluginName) {
       //get the config
```
JosefBredereck: This line could also be removed 👌

JosefBredereck (changes requested):
When those 3 things are resolved or clearified. I'm appy to approve this PR.

packages/core/src/index.js
JosefBredereck: `//get the config`

JosefBredereck:
Also, you will need to update the branch from `dev` or the merging will be blocked.

ringods:
@JosefBredereck will process your requests next week. I'm on holiday this week. ;-)

packages/core/src/lib/resolver.js
ringods: Patternlab loads files from installed NodeJS packages. It starts loading via whatever resolver a package manager could configure inside the Node runtime. In my case, I want to support Yarn v2 with PnP active eventually. In such a setup, no `node_modules` folder is used at all and every package is stored as a zip file. Here is an example how files inside a package are to be found:

https://yarnpkg.com/features/pnp#packages-are-stored-inside-zip-archives-how-can-i-access-their-files

packages/core/src/lib/loaduikits.js
ringods: `packages/core/test/util/patternlab-config.json` is left as is intentionally because it is used in the tests for checking the warning messages.

Question back: `packages/development-edition-engine-react/patternlab-config.json` doesn't have a `uikits` block, but does point to `./node_modules/@pattern-lab/uikit-workshop` files inside the `paths` section. With my change, this won't work in Yarn v2. I don't know if this is anything specific to React but I would need some help here on what to do.

JosefBredereck: Leave it out then. I'm not shure which engines are used anymore. Some feel depreacated / outdated anyway.

packages/core/src/lib/resolver.js
JosefBredereck: If it does not interfere with other resolvers like npm or yarn v1 everything is fine.

ringods: @JosefBredereck if uikit resolution works via `require.resolve`, the updated code *ought* to work under any other setup. If it doesn't it is a bug and needs to be fixed, which I'm glad to provide a fix for.

ringods:
@JosefBredereck can you review again? Please focus on the `uikits` loading for now. I will submit a new PR to update:

- `core/src/lib/starterkit_manager.js`
- `core/src/lib/pattern_engines.js`

so these also use the new resolver instead of calculating paths into `node_modules` themselves.

JosefBredereck (approved):
Good changes. Also, I like that you keep the old way in place and notify the user to update the config.

PR #1232, opened 2020-08-11
Fix default behavior of nested patterns

JosefBredereck:
Summary of changes:

Switched the default pattern creation way. For missing `.md` files there was an issue with nested pattern folders and also variants were not working as expected.


packages/core/test/object_factory_tests.js
```diff
@@ -117,8 +117,8 @@ tap.test('test Pattern name for variants correctly initialzed', function(test) {
       d: 123,
     }
   );
-  test.equals(p1.name, '00-atoms-00-global-00-colors-variant');
-  test.equals(p2.name, '00-atoms-00-global-00-colors-variant-minus');
+  test.equals(p1.name, '00-atoms-00-global-colors-variant');
+  test.equals(p2.name, '00-atoms-00-global-colors-variant-minus');
```
sghoweri, line 121: So glad to see this going away -- this has always been one of those annoying things with how the filename / URL structure was assembled 👍

packages/core/src/lib/object_factory.js
```diff
@@ -41,7 +41,10 @@ const Pattern = function(
   const info = this.getPatternInfo(
     pathObj,
     patternlab,
-    isPromoteToFlatPatternRun
+    isPromoteToFlatPatternRun ||
+      (patternlab &&
+        patternlab.config &&
+        patternlab.config.allPatternsAreDeeplyNested)
```
sghoweri, line 47: @JosefBredereck I can't wait for when we can switch to using Node 14+ which now supports optional chaining natively, no Babel transpiling required!

packages/docs/src/docs/pattern-organization.md
```diff
@@ -59,3 +59,5 @@ The `deeplyNested` attribute is used to toggle the pattern building behavior and
 
 - **deeplyNested not set or false**  - Pattern won't be handled as a deeply nested pattern
 - **deeplyNested: true**  - Pattern will be handled like mentioned under [Deeper Nesting](#heading-deeper-nesting)
+
+To turn on this behavior globally, just add `"allPatternsAreDeeplyNested": true` to your `patternlab-config.json`.
```
sghoweri, line 63: +1 so existing PL installs not already switched to using `allPatternsAreDeeplyNested` won't run into any file name / path changes; just applies to the new spec we're moving towards.

packages/core/src/lib/object_factory.js
JosefBredereck: Yes, I was using `patternlab?.config?.allPatternsAreDeeplyNested` first, but then it did not work. (Had still V13.x)

But we need to be careful since not everyone uses V14+ now and maybe not for a long time.

sghoweri (approved):
I'm good with this -- all small sensible changes + only scoped to the latest (new) opt-in `allPatternsAreDeeplyNested` config option. 👍 

Thanks @JosefBredereck!

sghoweri (changes requested):
Err - hold on a sec. Seeing something weird with the handlebars deployment

sghoweri (approved):
Ok - after merging in the reverted code off of `dev` the handlebars demo is once again working as expected. Phew 😅

PR #1233, opened 2020-08-16
Fix broken nav for too many entries

### LINKED ISSUE #1205: Cut off navigation when too many menu items
I am using Pattern Lab Node `v5.10.1` on `Windows`, with Node `v12.12.0`, using a `Gulp Mustache` Edition.

##### Expected Behavior
The Patternlab navigation should be fully usable when having many pages and beeing left aligned.

##### Actual Behavior
When the Patternlab navigation has many entries (27 of top and sublevel in my case), the content is cut off. The problem comes through the css rule: 
`.pl-is-active + .pl-js-acc-panel {
    max-height: calc(100vh - 2rem - 2rem);
}`

image

##### Steps to Reproduce
Have a left sided patternlab navigation where the numer of top or second level entries exceeds the screen height.

JosefBredereck:
Closes #1205

Some scss changes in the navigation to fix the issue.

### DIFF
--- a/packages/uikit-workshop/src/sass/scss/05-themes/_sidebar-theme.scss
+++ b/packages/uikit-workshop/src/sass/scss/05-themes/_sidebar-theme.scss
@@ -8,12 +8,12 @@
     */
   .pl-c-body--theme-sidebar {
     /**
-        * Header
-        * 1) Set width to sidebar width defined above
-        * 2) Make header 100% of the viewport height
-        * 3) Stack header items on top of each other
-        * 4) void bottom border for light theme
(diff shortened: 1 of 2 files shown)


sghoweri (approved):
1. Thanks for running these through Prettier! I feel like that was a quirky thing inherited from the earlier  Pattern Lab UI a lot of these files originated from... 😅 

2. 🤦 Man, this was one of those bugs I had already ran into and fixed in the Bolt Design System, but never managed to get the upstream fix PR'd back down before switching jobs.. 

image

PR #1236, opened 2020-08-16
Fix too much recursion and update node-sass

JosefBredereck:
Summary of changes:

When the UI-kit has been rendered there was a reappearing "too much recursion" error. That should be resolved with this PR.


ringods:
@JosefBredereck LGTM

sghoweri (approved):
Ah, yep. I'm super familiar with this `self` to `this` web component constructor updates -- all sensible changes and everything still seems to work fine when spot checking these particular components getting the updates.

Thanks @JosefBredereck!

PR #1240, opened 2020-08-19
load_plugin and is_plugin is not used anymore in the code base

ringods:
Summary of changes:
* `load_plugin` and `is_plugin` on `plugin_manager` is not used anymore in the codebase. Cleaning up.


JosefBredereck (approved):
Ah yes, the good old plugin code to load by name ^plugin- will now be gone for all time 👍

PR #1252, opened 2020-08-29
BREAKING-CHANGE: Deprecate ordering by file-prefix ##-

JosefBredereck:
Closes #1028
Closes #1176

Summary of changes:

Remove the ability to sort files and folders by the ##- prefix. Since that approach could become unhandy when you want to reorder your patterns.

(Reopening to split changes in code and starter kit file changes)

PR #1237, opened 2020-08-16
BREAKING-CHANGE: Deprecate underscore prefix

JosefBredereck:
Closes #1209 
Closes #1182 
Closes #1183 

Summary of changes:

Change the way pattern lab is hiding patterns. Because all those file system hacks are not a good approach we decided to remove this functionality in favor of a more flexible way.

PR #1253, opened 2020-08-29
Update starter kit file structure to match deprecation tickets

JosefBredereck:
To split up changes in #1252 and #1237 and the starter kit file changes I create a separate ticket that depends on the other two tickets.

I did this split to reduce the complexity of both PR's and resulting merge conflicts in the future.


JosefBredereck:
Should be merged with the two mentioned tickets.

PR #1255, opened 2020-09-01
Fix IE11 error - SCRIPT438: Object doesn't support property or method 'forEach'

shaal:
Running PL 5.13.0 on Internet Explorer 11, and I am getting this error.
```SCRIPT438: Object doesn't support property or method 'forEach'```

Summary of changes:
When this class is not in a page `'.pl-js-pattern-extra-toggle'`, the array is empty, and `forEach` throw an error.
Checking that the array is not empty before running forEach function.

### CODE REVIEW

File: packages/uikit-workshop/src/scripts/components/modal-styleguide.js [hunk 1]
```diff
@@ -22,12 +22,14 @@ export const modalStyleguide = {
     // go through the panel toggles and add click event to the pattern extra toggle button
     const toggles = document.querySelectorAll('.pl-js-pattern-extra-toggle');
 
-    toggles.forEach(toggle => {
-      toggle.addEventListener('click', e => {
-        const patternPartial = toggle.getAttribute('data-patternpartial');
-        modalStyleguide.toggle(patternPartial);
+    if (toggles.length > 0) {
```
ringods: @shaal could you replace the `forEach` completely with a `for` loop by index? According the docs, `querySelectorAll` returns a `NodeList` object and can be accessed by index.

https://www.w3schools.com/js/js_htmldom_nodelist.asp

See the example on iterating in the web page above.

ringods (changes requested):

JosefBredereck:
I would love to see pattern lab running in IE11. I had some issues with that for a customer.

sghoweri:
FWIW, the overhauled Pattern Lab UI previously was working ok (albeit a
little sluggish at times + with no CSS custom vars support) in IE 11...

Looks like the main UI is rendering (doing a super quick spot check) so
hopefully there's just a few bugs to smooth over:
https://www.webpagetest.org/result/200902_M8_2e9cad478c95bbfc70057edddc4a674f/1/screen_shot/#step_1

JosefBredereck:
Sorry that I highjacked this PR but there was much more to do then just adding an if. 
I also remove unused code in this case and uninstalled packages that are not in use anymore.

packages/uikit-workshop/src/scripts/components/modal-styleguide.js
ringods: @JosefBredereck even if you overhauled this PR, I think my remark still stands that it would be safer to loop over the `NodeList` object by index without needing the additional surronding `if` statement.

ringods (changes requested):

packages/uikit-workshop/src/scripts/components/modal-styleguide.js
JosefBredereck: As you see in the code I also changed that.

ringods (approved):

packages/uikit-workshop/src/scripts/components/modal-styleguide.js
ringods: I approved the PR. But given the condition of the `for` loop, I think the surrounding `if` statement is not needed anymore and can be removed.

shaal:
@JosefBredereck @ringods with the new syntax of the `for` loop, the `if` statement above it should be removed.
I wish I knew how to download this code, and test it myself (as oppose to adding PL official packages in package.json file)

packages/uikit-workshop/src/scripts/components/modal-styleguide.js
```diff
@@ -22,12 +22,14 @@ export const modalStyleguide = {
     // go through the panel toggles and add click event to the pattern extra toggle button
     const toggles = document.querySelectorAll('.pl-js-pattern-extra-toggle');
 
-    toggles.forEach(toggle => {
-      toggle.addEventListener('click', e => {
-        const patternPartial = toggle.getAttribute('data-patternpartial');
-        modalStyleguide.toggle(patternPartial);
-      });
-    });
+    if (toggles.length > 0) {
```
shaal: Now that we have an updated `for` loop syntax, the `if` statement above it should be remove.

packages/uikit-workshop/src/scripts/components/modal-styleguide.js
```diff
@@ -22,12 +22,14 @@ export const modalStyleguide = {
...
+      for (let i = 0; i < toggles.length; i++) {
+        toggles[i].addEventListener('click', e => {
+          const patternPartial = toggles[i].getAttribute('data-patternpartial');
+          modalStyleguide.toggle(patternPartial);
+        });
+      }
```
JosefBredereck:
```suggestion
    for (let i = 0; i < toggles.length; i++) {
      toggles[i].addEventListener('click', e => {
         const patternPartial = toggles[i].getAttribute('data-patternpartial');
        modalStyleguide.toggle(patternPartial);
      });
```

packages/uikit-workshop/src/scripts/components/modal-styleguide.js
JosefBredereck: done

JosefBredereck (approved):

JosefBredereck:
I just tested all changes; with the production source, we had so many problems in the past. Everything works fine in IE11 now. I will merge now.

JosefBredereck:
Released with [v5.13.1](https://github.com/pattern-lab/patternlab-node/releases/tag/v5.13.1)

JosefBredereck:
Released with [v5.13.1](https://github.com/pattern-lab/patternlab-node/releases/tag/v5.13.1)

PR #1256, opened 2020-09-01
Pattern Lab only loads the pattern engines from the config

### LINKED ISSUE #1251: Request for input: make pattern engine loading explicit via the PL config
This issue is to discuss the changes in loading pattern engines. It requires input from the core team: @bmuenzenmeyer @bradfrost @sghoweri @JosefBredereck and whoever I am forgetting.

In #1225 & #1246, I changed the way how resources from UIKits are loaded: via the NodeJS resolver (`require.resolve`). For this to work, it now requires the property `package`, with a fallback at the moment based on `name` but which will be removed in the future. This makes it the loading explicit: first add the UIKit as a package dependency, then resolve the resources and copy them. Using the resolver also makes this feature working independent of the package manager.

The next piece of code loading that needs to go via the resolver is the loading of pattern engines. The current code is again scanning the `node_modules` folder in 2 different locations manually:

https://github.com/pattern-lab/patternlab-node/blob/931b390af8abb7c9e6428d4fa27262d0af8b9dd2/packages/core/src/lib/pattern_engines.js#L12-L21

It returns every package which matches `engine-` in the package name:

https://github.com/pattern-lab/patternlab-node/blob/931b390af8abb7c9e6428d4fa27262d0af8b9dd2/packages/core/src/lib/pattern_engines.js#L23-L37

Like UIKits, this is not explicit and can lead to incorrect packages being found. If we want to make it explicit, the PatternLab configuration needs to be extended with an explicit mention of the pattern engines which must be loaded. But how should this look?

* Should the config only specify a single pattern engine or should it be a list?
* For each of the pattern engines, is more config needed besides the package name?

Any further suggestions?

ringods:
Closes #1251 

### Summary of changes:
* Load the engines which are configured in the config first
* Fallback to the old way of scanning `node_modules` if no engines are configured
* Documentation added on the `engines` property in the Pattern Lab config file. URLs in the source code should point to this property once it is published during the release.
* Proper deprecation messages should be in place.


JosefBredereck:
I can not approve that PR at the moment. The preview is not building.

Pattern Engine undefined: good to go
Found plugin: @pattern-lab/plugin-tab
Attempting to load and initialize plugin.
where is the engine?
Error rendering pattern lab header
Omitting atoms-colors from styleguide patterns because it has an underscore prefix.
where is the engine?
error during header render()
where is the engine?
error during footer render()
where is the engine?
Error building buildFooterHTML

JosefBredereck:
When I add engine mustache, it is working. (We need to get rid of that engine to be required everywhere but it's part of the UI-kit) But it still messages:
Pattern Engine _undefined_: good to go
Pattern Engine _undefined_: good to go

```
[...]
    "mustache": {
      "package": "@pattern-lab/engine-mustache",
      "fileExtensions": [
        "mustache"
      ]
    }
[...]
```

mfranzke:
@ringods first of all thanks a lot for your work that you did so far regarding this topic. I tried to follow up on the feedback by @JosefBredereck and checked in some minor changes of how I would think about solving this topic. I hope that was in your interest.

I've integrated some workarounds, that we should be able to remove with https://github.com/pattern-lab/patternlab-node/pull/1455; as even also mentioned by @JosefBredereck we should get rid of mustache in favour of its successor handlebars anyhow.

packages/docs/src/docs/advanced-config-options.md
```diff
@@ -402,7 +433,7 @@ Important details:
...
 - the [default `paths.public` object paths](https://github.com/pattern-lab/patternlab-node/pull/840/commits/812bab3659f504043e8b61b1dc1cdac71f248449) within `patternlab-config.json` are now relative to the current UIKit's `outputDir`. Absolute paths will no longer work. Someone could test putting an absolute path in a UIKit `outputDir` property and see what happens I suppose.
 - `dependencyGraph.json` has moved to the project root rather than `public/` as we should only retain one
-- The lookup of the uikit by `name` is deprecated and the user will be notified of it. If the `package` property isn't defined, there is a default fallback lookup strategy where the value of `name` is tried as:
+- The lookup of the uikit by `name` is **deprecated** and will be removed in v6. The user will be notified of it. If the `package` property isn't defined, there is a default fallback lookup strategy where the value of `name` is tried as:
```
JosefBredereck: `[...] will be removed in v7 [...]` when we do a major update in January.

packages/core/src/lib/pattern_engines.js
```diff
@@ -86,52 +99,106 @@ const PatternEngines = Object.create({
...
+    // Try to load engines! We load the engines configured in patternlab-config.json
+    const enginesInConfig = findEnginesInConfig(patternLabConfig);
 
-      logger.debug(
-        `Loading engines from ${engineDirectory.displayName}: ${engineDirectory.path} ...`
-      );
+    if (enginesInConfig) {
+      // Quick fix until we've removed @pattern-lab/engine-mustache, starting with https://github.com/pattern-lab/patternlab-node/issues/1239
+      // @TODO: Remove after removing @pattern-lab/engine-mustache dependency
+      enginesInConfig.mustache = enginesInConfig.mustache || {};
```
JosefBredereck, line 108: Interesting approach. Do we need to change this to handlebars after you changed the package with #1456?

JosefBredereck (changes requested):

packages/docs/src/docs/advanced-config-options.md
mfranzke: changed that.

packages/core/src/lib/pattern_engines.js
mfranzke: It's mainly a mitigation, because we still need the mustache engine as we reference mustache templates all over the place, and without it we would get errors like that the header isn't available. By https://github.com/pattern-lab/patternlab-node/pull/1455 we should be able to remove this code part again.

packages/docs/src/docs/advanced-config-options.md
JosefBredereck:
```suggestion
- The lookup of the uikit by `name` is **deprecated** and will be removed in v7. The user will be notified of it. If the `package` property isn't defined, there is a default fallback lookup strategy where the value of `name` is tried as:
```

JosefBredereck (approved):

PR #1287, opened 2021-02-05
docs(serverOptions): moved options from archived doc to the docs section

mfranzke:
Summary of changes:
Moved options settings for the live-server from archived document to the pages docs section, to include those options more prominently.


packages/docs/src/docs/advanced-config-options.md
```diff
@@ -220,7 +220,20 @@ Used to activate rendering flat patterns on view all pages and generate view all
...
+Sets [live-server options](https://github.com/pattern-lab/live-server#usage-from-node):
+
+| key        | example value                          | description                                                                                        |
+|------------|----------------------------------------|----------------------------------------------------------------------------------------------------|
+| port       | 8181                                   | Set the server port. Defaults to 8080.                                                             |
+| host       | "0.0.0.0"                              | Set the address to bind to. Defaults to 0.0.0.0 or process.env.IP.                                 |
+| root       | "/public"                              | Set root directory that's being served. Defaults to cwd.                                           |
+| open       | false                                  | When false, it won't load your browser by default.                                                 |
+| ignore     | 'scss,my/templates'                    | Comma-separated string for paths to ignore.                                                        |
```
JosefBredereck: This needs to be `ignorePattern`. `ignore` will overwrite the default not to watch `public.root` aka `public/`

mfranzke:
@JosefBredereck I‘ve copied the ones by the linked page but even also just realized that they don‘t match the implementation (any more).

packages/docs/src/docs/advanced-config-options.md
mfranzke: @JosefBredereck good catch. I‘ve copied these out of the linked page but even also just realized that they don‘t match the implementation (any more).

JosefBredereck: Can you adjust this?

mfranzke: Yes, sure.

mfranzke: @JosefBredereck I’m wondering about the best approach here.

I will insert ignorePattern. But regarding „ignore“: As it’s actually a provided option, should we either leave it out or just describe it correctly and warn about any overwriting? I tend to do the latter, to ensure completeness, avoid surprises and give those opportunities, that would like to tweak things to the next level.

What do you think?

JosefBredereck (approved):

PR #1303, opened 2021-03-20
refactor: updated further dependencies

mfranzke:
#### Simply updated
- [underscore](https://underscorejs.org/#changelog)(mainly because of [Vulnerability "Arbitrary Code Injection"](https://snyk.io/vuln/SNYK-JS-UNDERSCORE-1080984))
- [js-beautify](https://github.com/beautify-web/js-beautify/blob/main/CHANGELOG.md)
- [update-notifier](https://github.com/yeoman/update-notifier/releases)
- [portfinder](https://github.com/http-party/node-portfinder/releases)
- [hasha](https://github.com/sindresorhus/hasha/releases)
- [wolfy87-eventemitter](https://github.com/Olical/EventEmitter/blob/master/CHANGES.md)
- [scriptjs](https://github.com/ded/script.js/commits/master)
- [pwa-helpers](https://github.com/Polymer/pwa-helpers/commits/master)
- [mousetrap](https://github.com/ccampbell/mousetrap/releases)
- [clipboard](https://github.com/zenorocha/clipboard.js/releases)
- [@webcomponents/custom-elements](https://github.com/webcomponents/polyfills/blob/master/packages/custom-elements/CHANGELOG.md)
- [@webcomponents/webcomponentsjs](https://github.com/webcomponents/polyfills/blob/master/packages/webcomponentsjs/CHANGELOG.md)
- [get-own-property-symbols](https://github.com/es-shims/get-own-property-symbols/commits/v0.9.5)
- [lodash](https://github.com/lodash/lodash/commits/4.17.21)
- [cors](https://github.com/expressjs/cors/commits/v2.8.5?before=9158a8686d64bf567440d030873378c429ad60b0+35&branch=v2.8.5)
- [twing](https://github.com/NightlyCommit/twing/blob/master/CHANGELOG)
- [nunjucks](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md)
- [minimist](https://github.com/substack/minimist/commits/1.2.5)
- [glob](https://github.com/isaacs/node-glob/commits/v7.1.6)
- [jshint](https://github.com/jshint/jshint/blob/master/CHANGELOG.md)
- [morgan](https://github.com/expressjs/morgan/releases)
- [connect](https://github.com/senchalabs/connect/commits/3.7.0?before=fa8916e6350e01262e86ccee82f490c65e04c728+35&branch=3.7.0)
- [colors](https://github.com/Marak/colors.js/commits/v1.4.0)
[...]

PR #1301, opened 2021-03-15
refactor: updated devDependencies regarding webpack / to webpack version 5 & some cleanup

mfranzke:
Updated webpack to version 5, and all it's related dependencies

As I'm not that kind of a webpack user, I would probably link some of the necessary code changes to some older commits with which the code was introduced.

#### Simply updated
- [clean-css-loader](https://github.com/retyui/clean-css-loader/blob/master/CHANGELOG.md)
  - [x] major release update relates to [dropping webpack@1 support](https://github.com/retyui/clean-css-loader/blob/master/CHANGELOG.md#400)
- [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin/blob/main/CHANGELOG.md)
- ~~[node-sass-selector-importer](https://github.com/maoberlehner/node-sass-magic-importer/releases)~~ (got removed by #1373)
- [style-loader](https://github.com/webpack-contrib/style-loader/blob/master/CHANGELOG.md)
  - [x] major release is related to ["minimum supported `Node.js` version is `12.13.0`" and "minimum supported `webpack` version is `5.0.0`"](https://github.com/webpack-contrib/style-loader/releases/tag/v3.0.0)
- [svg-sprite-loader](https://github.com/JetBrains/svg-sprite-loader/blob/master/CHANGELOG.md)
- [svg-transform-loader](https://github.com/JetBrains/svg-mixer/blob/master/packages/svg-transform-loader/CHANGELOG.md)
- [svgo-loader](https://github.com/svg/svgo-loader/releases)
  - [x] Switched to SVGO 3, see its breaking changes: https://github.com/svg/svgo/releases/tag/v3.0.0
    - [x] Node.js 14+ is required for version
    - [x] stable package is replaced with native stable sort (required node 12+)
  - [x] Removed support of Webpack 4
- [ts-loader](https://github.com/TypeStrong/ts-loader/blob/master/CHANGELOG.md)
  - [x] major release update relates to ["minimum webpack version: 5" and "minimum node version: 12"](https://github.com/TypeStrong/ts-loader/blob/main/CHANGELOG.md#v900)
- ~~[no-emit-webpack-plugin](https://github.com/xipasduarte/no-emit-webpack-plugin/releases)~~ (this got removed by #1349)
- [babel-loader](https://github.com/babel/babel-loader/releases)
  - [x] major update is related to [Requiring babel ^7.12.0 and Node.js >= 14.15.0](https://github.com/babel/babel-loader/releases/tag/v9.0.0)
[...]

PR #1306, opened 2021-03-28
refactor: updated eslint devDepencencies

mfranzke:
### Summary of changes:

**It's important that someone else with ESLint, babel and webpack knowledge would have a look at these changes, especially my three code comments and the necessary evaluations.**

- Updated all eslint and related dependencies to newer versions and
- extracted the eslintconfig within the `live-server` package to an external config file

#### added

- [eslint-plugin-node](https://www.npmjs.com/package/eslint-plugin-node):
  - Two of the previously introduced eslint rules have been moved to this plugin by ESLint v7.0.0:
    - `no-mixed-requires` -> `node/no-mixed-requires`
    - `no-process-exit` -> `node/no-process-exit`
- [@babel/eslint-plugin](https://www.npmjs.com/package/@babel/eslint-plugin)
  > Companion rules for `@babel/eslint-parser`. `@babel/eslint-parser` does a great job at adapting `eslint` for use with Babel, but it can't change the built-in rules to support experimental features. `@babel/eslint-plugin` re-implements problematic rules so they do not give false positives or negatives.

#### updated dependencies

- [eslint](https://github.com/eslint/eslint/releases?after=v7.3.1)
- ~~[eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md)~~
- ~~[eslint-config-airbnb-base](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/CHANGELOG.md)~~
- [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier/blob/master/CHANGELOG.md)
  - [x] _This breaking change drops support for old versions of ESLint, Prettier and Node. You must use at least ESLint v7.28.0, Prettier v2.0.0 and Node v12.0.0. Aside from that, usage of this plugin remains identical._
  We're matching the ESLint version with the updates from this pull request and Prettier even already, as well as Node.
- [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md)

#### replaced

- "[`babel-eslint`](https://www.npmjs.com/package/babel-eslint) is now [`@babel/eslint-parser`](https://www.npmjs.com/package/@babel/eslint-parser).", source https://www.npmjs.com/package/babel-eslint and https://babeljs.io/blog/2020/07/13/the-state-of-babel-eslint

##### `rules`

- Replaced deprecated and removed rules by`ESLint@7`: https://eslint.org/docs/rules/#deprecated (see `added` section above as well)

##### `extends`

- https://github.com/prettier/eslint-plugin-prettier#recommended-configuration:
  > This plugin ships with a `plugin:prettier/recommended` config that sets up both the [plugin](https://www.npmjs.com/package/eslint-plugin-prettier) and [`eslint-config-prettier`](https://www.npmjs.com/package/eslint-config-prettier) in one go.

  and read further on the section

  > Exactly what does plugin:prettier/recommended do? Well, this is what it expands to:

##### ESLint and babel configs

sub-packages ESLint config file with an extend to root isn't necessary, as ESLint would search for that config anyhow, but we need to reference the root `babel.config.js` in our monorepo, compare to e.g.
https://babeljs.io/docs/en/config-files#root-babelconfigjson-file and https://babeljs.io/docs/en/config-files#file-relative-configuration

#### pending

- ~Why do we have both a `.eslintrc.js` and a `.eslintrc.json` in the root?~

#updatingdependencies[bit](https://github.com/pattern-lab/patternlab-node/pull/1299)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1300)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1302)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1301)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1305)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1303)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1307)by[bit](https://github.com/pattern-lab/patternlab-node/pull/1320)


.eslintrc.js
```diff
@@ -6,7 +6,8 @@ module.exports = {
     es6: true,
     browser: true,
   },
-  parser: 'babel-eslint',
+  parser: '@babel/eslint-parser',
+  // TODO: Check for "Note: The parserOptions described in the official documentation are for the default parser and are not necessarily supported by @babel/eslint-parser. Please see the section directly below for supported parserOptions.": https://www.npmjs.com/package/@babel/eslint-parser
   parserOptions: {
```
mfranzke, line 11: Check for "Note: The `parserOptions` described in the official documentation are for the default parser and are not necessarily supported by `@babel/eslint-parser`. Please see the section directly below for supported parserOptions.": https://www.npmjs.com/package/@babel/eslint-parser

.eslintrc.js
```diff
@@ -17,8 +18,9 @@ module.exports = {
     },
   },
   globals: {},
-  plugins: ['prettier'],
-  extends: ['eslint-config-prettier'].map(require.resolve),
+  plugins: ['@babel', 'prettier'],
+  extends: ['plugin:prettier/recommended'],
+  // TODO: Check for "ESLint rules relating to code formatting", compare to "This plugin works best if you disable all other ESLint rules relating to code formatting, and only enable rules that detect potential bugs. (If another active ESLint rule disagrees with prettier about how code should be formatted, it will be impossible to avoid lint errors.)", source: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
   rules: {
```
mfranzke, line 24: Check for "ESLint rules relating to code formatting", compare to "This plugin works best if you disable all other ESLint rules relating to code formatting, and only enable rules that detect potential bugs. (If another active ESLint rule disagrees with prettier about how code should be formatted, it will be impossible to avoid lint errors.) You can use [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to disable all formatting-related ESLint rules.", source: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration

packages/uikit-workshop/webpack.config.js
```diff
@@ -78,6 +78,7 @@ module.exports = function(apiConfig) {
           ],
         ],
         plugins: [
```
mfranzke, line 76: From what I understand, we should even also be able to move those to the `babel.config.js` in root, to align both linting and building here

JosefBredereck:
Somehow this one does not want to deploy to our preview systems 😟

mfranzke (commented):
> Somehow this one does not want to deploy to our preview systems 😟

Yeah, I‘ve sadly recognized this as well, and don‘t understand the reason so far after having a quick look. I will probably have to investigate further.

@JosefBredereck, who has access to the logs and could probably shed some light on what‘s happening there and any error log messages ?

JosefBredereck:
> who has access to the logs and could probably shed some light on what‘s happening there and any error log messages?

I think @sghoweri has access

packages/core/test/files/_meta/_head.html
```diff
@@ -20,4 +20,5 @@
     {{{ patternLabHead }}}
     <!-- End Pattern Lab -->
   </head>
-  <body class="{{ bodyClass }}">
+  <body class="{{ bodyClass }}"></body>
+</html>
```
mfranzke, line 24: Need to revert the closing html tag (in further templates here as well).

packages/core/src/lib/server.js
```diff
@@ -36,30 +36,6 @@ const server = (patternlab) => {
             patternlab.config.paths.public.root
           )
         );
-        defaults.assets = [
```
JosefBredereck, line 39: I know this is an old PR which is still WIP, but could you elaborate why this gets removed when the PR only is about ESLint changes?

packages/edition-node-gulp/gulpfile.js
```diff
@@ -27,7 +27,6 @@ function serve() {
   return patternlab.server
     .serve({
       cleanPublic: config.cleanPublic,
-      watch: true,
```
JosefBredereck, line 30: Same goes for the removal of this line

PR #1311, opened 2021-04-13
refactor(nestedpatterns): trigger warning message only after checking for whether it's actually a pattern file

### LINKED ISSUE #1309: Don't trigger deep pattern warning on non-pattern files
I am using Pattern Lab Node `v5.13.2` on `Linux`, with Node `v14.16.0`, using a `Custom` Edition.

##### Expected Behavior

This is regarding the warning that's emitted if you go beyond 3 levels deep when [deeper nesting](https://patternlab.io/docs/overview-of-patterns/#heading-deeper-nesting)'s enabled.

We *don't* have any patterns deeper than 3 levels, but we do store images in a subdirectory. That is, the structure is similar to

```
- molecules/
  - blocks/
    - media-block/
      - media-block.md
      - media-block.mustache
      - media-block.scss
      - media-block~variant1.json
      - media-block~variant2.json
        - images/
          - image1.jpg
```

I'd expect this to compile without warnings because there are no _patterns_ nested deeper than 3 levels.

##### Actual Behavior

The following warning's emitted.

```
Warning:
A pattern file: xxx was found greater than 3 levels deep from yyy.
It's strongly suggested to not deviate from the following structure under _patterns/
[patternType]/[patternSubtype]/[patternName].[patternExtension]
or
[patternType]/[patternSubtype]/[patternName]/[patternName].[patternExtension]

While Pattern Lab may still function, assets may 404 and frontend links may break. Consider yourself warned. 
Read More: https://patternlab.io/docs/overview-of-patterns/
```

Is there actually any risk with our current setup? Does it make sense to explicitly look for patterns rather than _any_ file before issuing the warning? If that's not workable for some reason (and assuming we're not doing anything dangerous) would it be possible to have an option to disable the warning? Currently we've increased the log level (as done in https://github.com/pattern-lab/patternlab-node/issues/1037) but ideally we wouldn't have to.

##### Steps to Reproduce

1. Set `allPatternsAreDeeplyNested` to `true` in `patternlab-config.json`.
2. Set up a pattern with the structure shown above.
3. Build pattern lab.

(And thanks for a fantastic project!)

mfranzke:
Closes #1309 

Summary of changes:
To ensure that the warning message on patterns nested too deep in the folder structure is only shown if the files stored in those subfolders are actually patterns (as the warning message describes them), we have to put the method for checking for whether the current file is a pattern before triggering that warning message.


andriokha:
:+1: Thanks @mfranzke, appreciated! I can confirm this fixes the problem for me as described in #1309.

JosefBredereck (approved):

PR #1334, opened 2021-06-19
feat(documentation): added (sub)groups documentation again #1262

### LINKED ISSUE #1262: Add documentation to pattern group and subgroup
In the code, the snipped at the end could be found. This behavior is not working as expected and will lead to issues following in our implementation. The intention was to document pattern groups and subgroups. We have no way to show this documentation, and also the group or subgroup will not be rendered when higher order documentation is available.

We will need a new approach to this special case.
First of all, we need to implement something similar to the pattern documentation but just for the groups and subgroups. Afterward, we need to implement a method to gather this data and place it in the correct position.

For not the code snipped will be removed in #1252 and can be implemented again.

```javascript
// loadPattern.js : 55
const mp = require('./markdown_parser');
  const markdown_parser = new mp();
  if (fileObject.ext === '.md') {
    try {
      const proposedDirectory = path.resolve(
        patternlab.config.paths.source.patterns,
        fileObject.dir,
        fileObject.name
      );
      const proposedDirectoryStats = fs.statSync(proposedDirectory);
      if (proposedDirectoryStats.isDirectory()) {
        const subgroupMarkdownFileContents = fs.readFileSync(
          proposedDirectory + '.md',
          'utf8'
        );
        const subgroupMarkdown = markdown_parser.parse(
          subgroupMarkdownFileContents
        );
        const subgroupPattern = new Pattern(relPath, null, patternlab);
        subgroupPattern.patternSectionSubgroup = true;
        subgroupPattern.patternDesc = subgroupMarkdown
          ? subgroupMarkdown.markdown
          : '';
        subgroupPattern.flatPatternPath =
          subgroupPattern.flatPatternPath + '-' + subgroupPattern.fileName;
        subgroupPattern.isPattern = false;
        subgroupPattern.engine = null;
        patternlab.subgroupPatterns[
          subgroupPattern.patternPartial
        ] = subgroupPattern;

        return subgroupPattern;
      }
    } catch (err) {
      // no file exists, meaning it's a pattern markdown file
      if (err.code !== 'ENOENT') {
        logger.warning(err);
      }
    }
  }
```

mfranzke:
Closes #1262

### Summary of changes:
Readded the functionality of showing the documentation for subgroups again – having some documentation for the groups as well is being prepared by this, but probably didn't work previous to the deletion of the documentation of the subgroups by #1252 anyhow.

#### Solution
We use the markdown content provided within the (sub)groups companion `.md` files, which are even already [defined like this for the ordering](https://patternlab.io/docs/reorganizing-patterns/#heading-re-ordering-pattern-groups-and-subgroups).

That for [some demo content has been added to the buttons subgroup](https://github.com/pattern-lab/patternlab-node/blob/c12a4f4dd3f40a2bcecb7bcdaf99e6fbfdde668d/packages/starterkit-handlebars-demo/dist/_patterns/molecules/buttons/_buttons.md), compare to the [rendered result](https://deploy-preview-1334--patternlab-handlebars-preview.netlify.app/?p=viewall-molecules-buttons).

Additionally some [documentation has been adapted](https://deploy-preview-1334--patternlab-docs-preview.netlify.app/docs/documenting-patterns/#heading-documenting-subgroups) from the reorganizing patterns page.


packages/docs/src/docs/pattern-documenting.md
```diff
@@ -55,6 +55,22 @@ We'd name our documentation file:
...
+
+
+To document pattern subgroups, you need to create a companion `.md` file for that subgroup. For example create `_patters/atoms/buttons/_buttons.md` for a pattern subgroup. In the `.md` file, the above concept can be applied. The doc-file resolving works the following `{patternsRoot}/{pattern-group folder name}/{pattern-sub-group folder name}/_{pattern-sub-group raw name without prefixes}.md`
+
+
+```markdown
+---
+order: 0
+---
+
```
JosefBredereck: That part can be removed from this documentation since it is already part of [Re-ordering pattern groups and subgroups](https://patternlab.io/docs/reorganizing-patterns/#heading-re-ordering-pattern-groups-and-subgroups)

packages/core/src/lib/ui_builder.js
```diff
@@ -140,7 +140,9 @@ const ui_builder = function () {
         patternName: _.startCase(
           isSubgroupPattern ? pattern.patternSubgroup : pattern.patternGroup
         ),
-        patternDesc: '',
+        patternDesc: isSubgroupPattern
+          ? pattern.patternSubgroupData.markdown
+          : pattern.patternGroupData.markdown,
```
JosefBredereck, line 145: Was it really that simple? I know I removed a very complicated functionality from the code in the past, and here it looks easy 👍

JosefBredereck (commented):
As always, your Pull Request description is excellent 🥇

packages/core/src/lib/ui_builder.js
mfranzke: Well, I even also wondered about that … ;-) And I must admit that I even also needed to think about and find this place a little longer.

The previous `patternDesc: ''` on line 143 was mainly leading to that always empty pattern description that was introduced by the code provided with commit ede5839457cb630fd164742613eb7f74750f930a – I think that this original place was never really enhanced to a regular implementation, but mainly overlayed by the code parts that you previously removed. Nevertheless the code in `ui_builder.js` isn't just a skeleton but contains some more direct definitions like the `patternName`, so I think that its a perfect place to reintroduce it over here.

packages/docs/src/docs/pattern-documenting.md
mfranzke: I was wondering of whether to include at least the one (and only ?) attribute for subgroups in this example, but you're perfectly right – as this is even already part of that reordering aspect, I've removed it from here.

PR #1345, opened 2021-08-20
fix(docs): use "UIKits" instead of "StyleguideKits"

renestalder:
Closes –

Summary of changes:

Align the terminology of custom pattern lab UIs with what is used in the configuration currently.
The documentation now uses "UIKits" to align with the "uikits" property of the pattern lab
configuration.


JosefBredereck:
Thank you for this improvement 👍

PR #1351, opened 2021-09-05
refactor: preventing problem on nested PL output

mfranzke:
### Summary of changes:
On a patternlab installation nested in subfolder(s), the main link on the logo currently points to the top level root. By changing the default from `"/"` to `"./"` we could support both nested installations as well as the ones directly hosted in the root of the URL.


JosefBredereck:
Useful improvement 👍

JosefBredereck (approved):

PR #1358, opened 2021-10-05
refactor: slightly increased the minimum node version

mfranzke:
### Summary of changes:
As we haven't increased the minimum expected node version number in a while, it's time to at least push it by one minor version to satisfy the installation of packages like `@babel/eslint-parser`.


JosefBredereck (approved):

JosefBredereck:
Maybe #1306 runs smoothly after that update 🙏

PR #1367, opened 2021-11-02
Includes additional twig syntax featires

mogsie:
The twig tags "from," "import" and "use" are not understood by patternlab.  They all use the same techniques as include, embed etc when resolving template names.

* `from` is documented in [macros](https://twig.symfony.com/doc/2.x/tags/macro.html#importing-macros):
* [import](https://twig.symfony.com/doc/2.x/tags/import.html)
* [use](https://twig.symfony.com/doc/2.x/tags/use.html)

Some twig templates use  {%- ... -%} syntax that gobble up the whitespace on each side. This change ensures that this syntax is understood when looking for partials.

* [Whitespace control](https://twig.symfony.com/doc/2.x/templates.html#whitespace-control)

I wasn't aware of the `{%~` syntax (line-whitespace trimming) introduced in 2.8; I could add support for that too.


JosefBredereck:
> Some twig templates use {%- ... -%} syntax that gobble up the whitespace on each side. This change ensures that this syntax is understood when looking for partials.

Will this break any existing functionality?

mogsie:
> Will this break any existing functionality?

No, it shouldn't.  This code is only used when determining the "graph" of template dependencies, as far a I know.  Twig understands the `{%- .. -%}` syntax, and has done so forever.  When you use `{%- include` in a twig file, patternlab doesn't detect the dependency between this template and the included template.

mogsie:
I noticed I changed "engine-twig" and not "engine-twig-php" which I linked for documentation purposes; I should probably do both, and align this properly.

JosefBredereck (approved):

PR #1368, opened 2021-11-03
fix(core): Subgroup cannot be hidden

### LINKED ISSUE #1366: Hiding pattern sub groups ignored
I am using Pattern Lab Node `v5.15.1` on `Linux`, with Node `v16`, using a `Twig` Edition.

##### Expected Behavior

If I add `components/buttons/_buttons.md` with the content:

```
---
hidden: true
---
```

That the "Buttons" group does not appear in the output of the documentation.

Relevant part of the patternlab documentation: https://patternlab.io/docs/hiding-patterns-in-the-navigation/#heading-hiding-pattern-sub-groups

##### Actual Behavior

The option seems to be ignored for a whole folder. At least for the Twig edition.

Hiding single files works as documented.

##### Steps to Reproduce

1. Create a sub group structure e.g. "components/buttons".
2. Add a test twig file to that folder, e.g. "components/buttons/test.twig" with some content.
3. This appears in the documentation under the menu "components" as "buttons" then "test".
4. Now add `components/buttons/_buttons.md` with the content above (as in the expected behaviour.
5. The components -> buttons group should now no longer appear in the documentation menu. For me, it still does

Simplified example.

JosefBredereck:
Closes #1366

### Summary of changes:

In the past, when I refactored the subgroup namings, I did miss those parameters. It is so trivial and took some time to find.

### DIFF
--- a/packages/core/src/lib/ui_builder.js
+++ b/packages/core/src/lib/ui_builder.js
@@ -102,7 +102,7 @@ const ui_builder = function () {
     // this pattern is contained with a directory documented as hidden (a handy way to hide whole directories from the nav
     isOmitted =
       (pattern.patternGroupData && pattern.patternGroupData.hidden) ||
-      (pattern.patternSubGroupData && pattern.patternSubGroupData.hidden) ||
+      (pattern.patternSubgroupData && pattern.patternSubgroupData.hidden) ||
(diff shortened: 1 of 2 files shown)


mfranzke (approved):
looks totally reasonable to me.

Issue #1380, opened 2021-11-29
Too many dev related listed as regular dependencies


mfranzke:
I am using Pattern Lab Node `v5.15.3` on `Mac`, with Node `v12.13.1`, using a `handlebars` Edition with the demo patterns installed.

##### Expected Behavior

All build, testing and release dependencies of the different packages are included as `devDependencies` within the `package.json` files and not as `dependencies`. This would both shrink the weight and amount of installed (and partly outdated) packages for consumers of the pattern lab packages.

##### Actual Behavior

Some of the dependencies are listed as `dependencies` instead of `devDependencies`.

##### Steps to Reproduce

Install pattern lab with `npm create pattern-lab` and have a look at the `node_modules` folder, that contains dependencies like webpack, that are only used during preparding the package `@pattern-lab/uikit-workshop` for publishing.

PR #1384, opened 2021-11-30
refactor: moved dependencies to devDependencies @ cli

mfranzke:
Solves parts of #1380

### Summary of changes:

Moving `dependencies` to `devDependencies` that are only used during preparding the package `@pattern-lab/uikit-workshop` for publishing.


packages/cli/package.json
```diff
@@ -24,18 +23,19 @@
     "lodash": "4.17.21",
     "ora": "2.1.0",
     "path-exists": "3.0.0",
-    "sanitize-filename": "1.6.3",
```
JosefBredereck, line 27: `sanitize-filename` is not in use anymore?

JosefBredereck (approved):

packages/cli/package.json
mfranzke: @JosefBredereck exactly. I've left it within this pull request to only make the move of the other dependencies, but will provide a new pull request with an explanation and code references that provide the reasons for its removal.

JosefBredereck (approved):

PR #1396, opened 2021-12-17
refactor: removed unused and unnecessary dependencies

mfranzke:
### Summary of changes
Removed the following dependencies, that were either unused or unnecessary:
- [`path-exists`](https://www.npmjs.com/package/path-exists):
  > NOTE: `fs.existsSync` has been un-deprecated in Node.js since 6.8.0. If you only need to check synchronously, this module is not needed.
  
  Source: https://www.npmjs.com/package/path-exists
- `@pattern-lab/live-server` in `@pattern-lab/cli`: This package is not used in the cli
- [`sanitize-filename`](https://www.npmjs.com/package/sanitize-filename): This package has been introduced to the cli package with commit a50fb3a3845246f828370c5bcb27a4a25ad858ec, but never been used
- [`eslint-plugin-import`](https://www.npmjs.com/package/eslint-plugin-import): This plugin doesn't seem to be in use, neither within the code of its [introduction commit](https://github.com/pattern-lab/patternlab-node/commit/5b7a057d46ccd16b5832af1441030c7b76f237a8) nor afterwards.
- [`eslint-config-airbnb-base`](https://www.npmjs.com/package/eslint-config-airbnb-base): This config doesn't seem to be in use, neither within the code of its [introduction commit](https://github.com/pattern-lab/patternlab-node/commit/5b7a057d46ccd16b5832af1441030c7b76f237a8) nor afterwards.


JosefBredereck:
I like that. So we lose some weight 🚀

JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.0

PR #1399, opened 2021-12-25
refactor: corrected some aspects out of eslint feedback

mfranzke:
### Summary of changes:
Fixed some of the aspects that comes out of eslint feedback (via `yarn lint`).


JosefBredereck:
Do we need the commented-out code, or can we remove it entirely?

mfranzke:
> Do we need the commented-out code, or can we remove it entirely?

@JosefBredereck I was struggling with this question as well. I decided to leave it in as we don't need it at the moment, but as soon as someone might need e.g. the `event` or `options` parameter, they could even already follow the naming (pattern) and don't need to look this up. But that's the only small reason that I came up with that pulled me into the "leave it in direction", whereas I'm fine with both ones, so I could still remove them. What do you think?

mfranzke:
> Do we need the commented-out code, or can we remove it entirely?

I've just decided to remove them. So we should be good to go now.

JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.0

PR #886, opened 2018-07-10
fix(package): Allow .json extension on annotations file (issue #836)

engelfrost:
Closes #836 

Summary of changes:
Read annotations.json if it exists. 
Else read annotations.js. 

Not sure if being able to read the old file is a necessary convenience. If not, that part of the code (the else statement) can easily be removed.

Issue #836, opened 2018-04-03
Rename annotations.js to annotations.json


henk23:
It's a json file, so why does it have a js file ending?
My IDE complains about corrupted js syntax and I think it is right about it.
What am I not seeing here?

bmuenzenmeyer:
@henk23 it's a good question!

what you aren't seeing is the support for the legacy Pattern Lab annotations file - which was a js file instead of JSON/YML.  Since we are in 3.0 Alpha mode anyways, it'd be smart to fix this once and for all, which will make the code simpler too!

bmuenzenmeyer:
fixed

PR #1402, opened 2022-01-02
refactor: allow .json extension on annotations file #836

mfranzke:
### Summary of changes:
As the [original implementation](https://github.com/pattern-lab/patternlab-node/pull/886) by @engelfrost (kudos for that!) accidentally got reverted with commit 9c49a093e5007b3992c042933031fed36b83b753, we should add that nice refactoring again to ensure that annotations JSON files could get used with the correct filetype / file-ending, and even also advocate for that file-ending from now on.

@engelfrost implementation even also ensured backwards compatibility, so this wouldn't be a breaking change.

The resulting implementation could get reviewed on https://deploy-preview-1402--patternlab-handlebars-preview.netlify.app/?p=viewall-molecules-global&view=annotations and https://deploy-preview-1402--patternlab-handlebars-preview.netlify.app/?p=pages-homepage&view=annotations

There's still a problem with displaying the tooltips on the ["standalone" view (in our example for the header)](https://deploy-preview-1402--patternlab-handlebars-preview.netlify.app/?p=pages-homepage&view=annotations), as the general styles from `pattern-lab.css` aren't included in this view.
This problem is unrelated to the fixes in this pull request and I've opened another pull request for that to fix: https://github.com/pattern-lab/patternlab-node/pull/1406


JosefBredereck:
How do you find stuff like this all the time? That's crazy 😄 
One commit with 253 changes, and you know what got removed by that commit.

mfranzke:
@JosefBredereck … oh, don't ask … ;-) others read novels, I read git commit history …

packages/core/src/lib/annotationExporter.js
```diff
@@ -6,39 +6,47 @@ const _ = require('lodash');
...
+        const jsPath = path.resolve(paths.source.annotations, 'annotations.js');
+
+        annotations = fs
+          .readFileSync(jsPath, 'utf8')
+          .replace(/^\s*var comments ?= ?/, '')
+          .replace(/};\s*$/, '}');
+
+        logger.info(
+          `Please convert ${jsPath} to JSON and rename it annotations.json.`
+        );
```
JosefBredereck, line 36: I'm a bit on the fence with that. We should let the user decide if he wants to use JS or JSON, even in the future.
Furthermore, we could check if it is possible to `require` that `.js` or `.json` file so that we do not have to strip comments and anything else in that file. Maybe we can change the way it needs to be written to a `module.exports` format.

mfranzke: @JosefBredereck, actually I've realized that we're stripping that `var comments` part first of all, to add it again on exporting the data: https://github.com/pattern-lab/patternlab-node/blob/41b4ec15243ba8b1e56bb1efce7d39dfe4c4e150/packages/core/src/lib/exportData.js#L72

So another solution might be to skip that whole transition to a pure JSON file and only provide the possibility with using JS (the existing format), but remove the JSON file possibility, as we're even already providing another format with the [markdown notation](https://patternlab.io/docs/adding-annotations/#heading-markdown-example) as well.
So another task for this ticket could be still to correct the documentation part anyhow: https://patternlab.io/docs/adding-annotations/#heading-json-example

What do you think?

JosefBredereck:
Another thing I just discovered and never understood: The annotations are written once and visible on all patterns there are available. Why should I want that?
Would it also be helpful to have a list like `useIn: [ <list of pattern names> ]`?

mfranzke:
> Another thing I just discovered and never understood: The annotations are written once and visible on all patterns there are available. Why should I want that?
> Would it also be helpful to have a list like `useIn: [ <list of pattern names> ]`?

well, they are as global as the global data, but there's no "scoped" or "local" annotations, related to a pattern even only. This even also made me wonder. Probably this would be a nice enhancement / future feature ?

JosefBredereck:
> > Another thing I just discovered and never understood: The annotations are written once and visible on all patterns there are available. Why should I want that?
> > Would it also be helpful to have a list like `useIn: [ <list of pattern names> ]`?
> 
> well, they are as global as the global data, but there's no "scoped" or "local" annotations, related to a pattern even only. This even also made me wonder.

Or I'm just blind. There is the `el: ...` part which defines to which elements these annotations get added. That was not clear to me 🥲

packages/core/src/lib/annotationExporter.js
JosefBredereck: No way 😄. In that case, throw away those nasty JSON files 🗑️

mfranzke: okay, sounds reasonable to me. I've set this ticket on draft again and will proceed by removing these parts as well as the nasty existing transitions in the codebase, as well as correct the documentation page.

JosefBredereck: The whole file you mentioned is a solution, but if we would rework the codebase with today's standards, we would maybe not do it that way.

mfranzke: so does it mean that you want to come up with (or at least describe) a solution ?

JosefBredereck: Not for now; it would require a more significant rewrite.
This line shows that we already work with modules somewhere.

https://github.com/pattern-lab/patternlab-node/blob/41b4ec15243ba8b1e56bb1efce7d39dfe4c4e150/packages/core/src/lib/exportData.js#L88

But then we also need to adjust the place where we load those annotations again.

What makes me curious. Annotations always have a comments property and no other. Is it possible that they have different properties? If not, this would be another change I would like to implement somehow.

https://github.com/pattern-lab/patternlab-node/blob/41b4ec15243ba8b1e56bb1efce7d39dfe4c4e150/packages/core/src/lib/exportData.js#L72

mfranzke: > No way 😄. In that case, throw away those nasty JSON files 🗑️

@JosefBredereck, on the other hand, besides the strange way to include the JSON data into the website after a transformation, the positive way of using JSON (even only) would be that this adapts the way other data is being maintained in the pattern lab system (like pattern data).

JosefBredereck: That is correct. Nodejs is so fast in transforming the data that we could use the approach that is already in this PR. In the end, it makes no difference.

JosefBredereck:
Some minor adjustment is standing out after the merge of #1406
https://github.com/pattern-lab/patternlab-node/pull/1406/files#diff-d562b65f94fc7de820db229a45973334e2c8195978fbc7643d1916ed6e63fd19

mfranzke:
> Some minor adjustment is standing out after the merge of #1406 https://github.com/pattern-lab/patternlab-node/pull/1406/files#diff-d562b65f94fc7de820db229a45973334e2c8195978fbc7643d1916ed6e63fd19

good catch, thanks. Just checked in the migrated file.

JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.0

PR #1403, opened 2022-01-02
chore: corrected some hyperlinks

mfranzke:
### Summary of changes:
Corrected some of the URLs within the code, either regarding
- that we should prefer HTTPS over HTTP nowadays and even also prevent the redirect
- that the URLs have changed


JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.0

PR #1405, opened 2022-01-02
refactor: always include dimensions on images

mfranzke:
We should advocate for good and performant HTML code, that should even already include dimensions for the browsers to be able to calculate it's dimensions and prevent recalculations of the design / redraw, page jumps, etc.

### Summary of changes:
Added dimensions to most of the images.


JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.0

PR #1406, opened 2022-01-02
fix(annotations): displaying annotation tooltips correctly

mfranzke:
We're currently facing problems especially on the ["standalone" view of patterns (e.g. `pages`)](https://deploy-preview-1402--patternlab-handlebars-preview.netlify.app/?p=pages-homepage&view=annotations), in which the general `pattern-lab.css` isn't referenced, that includes the `/packages/uikit-workshop/src/sass/scss/04-components/_annotations.scss` file.

### Summary of changes:
- Hiding elements (in this case the annotation tooltips) in a modern way by the `hidden` attribute instead of style property manipulations.
- Extracted parts of the annotations CSS code so that it could even also get imported into `pattern-scaffolding.css` (which we needed to convert to a SCSS file and transform that one as well).

### Preview URL with the fixed version
["standalone" view of patterns (e.g. `pages`)](https://deploy-preview-1406--patternlab-handlebars-preview.netlify.app/?p=pages-homepage&view=annotations)


JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.0

PR #1428, opened 2022-02-10
refactor: make removing index.html from URL configurable

### LINKED ISSUE #1427: No possibility to configure whether or not to remove index.html from the URL when adding to pushState
_inspired by and adapted from https://github.com/pattern-lab/styleguidekit-assets-default/pull/49/_

> We had a few issues, where:
> - Refreshing any page resulted in a 404 Page not found, either manually or after running php core/console --generate and reloading (we have a gulp task that watches for any changes in patterns, runs that command, and reloads the browser).
> - We couldn't share links between the team, since they returned 404s as well.

> We found that the cause was that Pattern Lab removes index.html from the URL, and our hosting environment doesn't load index.html by default.
> Our fix is to not remove index.html. If accepted, it wouldn't cause problems for anybody else and it would help PL to not assume the servers are configured to serve index.html as the default file.

I am using Pattern Lab Node `v5.16.2` on `Mac`, with Node `v12..13.1`, using a `Vanilla` Edition.

##### Expected Behavior

Be able to decide in whether my system redirects from **/index.html?** to **/?** (comparable to SPA) or not, as I might host it in an environment, which doesn't handle the default document of `index.html` correctly, like e.g. Gitlab Pages (it's in their backlog to support SPA properly, but it's resulting in a problem at the moment at least after a reload).

##### Actual Behavior

A URL containing the file `index.html` is always being redirected to the equal one with `index.html` being removed.

##### Steps to Reproduce

You need to reproduce this on an environment that doesn't handle the default file of `index.html` correctly in case of that no file is being provided at the URL. Most it works perfectly correctly on your localhost (http://localhost:3000/?p=all) and on our preview environment on Netlify.

mfranzke:
Closes #1427

### Summary of changes:
I've decided to make this change configurable, especially for backwards "compatibility" (or in other words not changing the current behavior).
Providing the possibility to deactivate the initial redirect and changing of the history towards the removal of an included `index.html` within the URL: from `/index.html?p=WHATEVER` to `/?p=WHATEVER`, as this might cause trouble on environments that don't acknowledge an `index.html` file if no file is provided in the URL.

Even though that this is a suboptimal configuration, under some circumstances you're able to change it (like on a public static hosting provider).

### DIFF
--- a/packages/docs/src/docs/advanced-config-options.md
+++ b/packages/docs/src/docs/advanced-config-options.md
@@ -113,6 +113,14 @@ This replaces the now obsolete `debug` flag.
 
 **default**: `info`
 
+## noIndexHtmlremoval (optional)
+
+You might host your pattern lab in an environment that doesn't acknowledge the default file to be `index.html` in case that none is provided – Gitlab Pages is an example for this, where you might publish your preview builds to during development. So in case that you get a URL like `<PATH>/index.html`, it will redirect to `<PATH>/?p=all`. You might not be able to share this resulting URL or even do a refresh of the page, as the server would respond with a 404 error.
+
+To disable this redirect which is similar to how Single Page Applications work, you could set the optional configuration `"noIndexHtmlremoval"` to `true`.
+
+**default**: `false`
+
 ## outputFileSuffixes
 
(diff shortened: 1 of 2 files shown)


JosefBredereck (approved):

JosefBredereck:
PR was released with v5.16.3

PR #1435, opened 2022-04-10
feat(engine-twig): add custom twing extensions

### LINKED ISSUE #1230: Adding custom TwingExtension functions and filters in engine-twig.
I am using Pattern Lab Node `v5.11.1` on `Linux`, with Node `v14.6.1`, using a Twig Edition.

Installed Dependencies:
    "@pattern-lab/cli": "^5.11.1",
    "@pattern-lab/core": "^5.11.1",
    "@pattern-lab/engine-twig": "^5.10.2",
    "@pattern-lab/uikit-workshop": "^5.11.1"

I currently am using the old Pattern Lab PHP with Twig engine at work for our design library, and am exploring what it would take to update pattern lab to the new versions in node.

My comment specifically addresses Twig Extensions for functions and filters.

In PHP PL with Twig, creating twig extensions is dirt simple, and we have many that we use in the course of our work.
When I initially tried out Node PL, I installed it with the included twig engine which was `engine-twig-php`.  I noticed there was a file in the root PL directory which was to be used to twig extensions.

I then read that, unless I misunderstood, the @patternlab/engine-twig should be the de facto twig engine in the node environment because of it's full-featured compatibility with the Twig Specs via twing, I reinstalled my PL Node environment to use that instead.

My issue is that after reading the files and documentation for engine-twig and twing in general, I can't seem to get any custom TwingExtension classes to work with pattern lab as the engine builds the template, leaving me with build errors stating custom filters or functions are unrecognized.

I don't know if I'm either missing something critical or if this isn't currently allowed in the `engine-twig` dependency as of yet.

If it is possible and someone could point me in the right direction, that would be appreciated.

If it is not currently a feature, I would like to request that the ability to pass in custom extensions to the renderer as it executes via a `customExtension.ts/js` file in the PL directory or something similar be added to Pattern Labs Twing Engine.

hadl:
Closes: #1230

Summary of changes:

 - patternlab-config.json new config key possible `engines -> twig -> loadExtensionsFile`
 - create JS file in PatternLab root directory (e.g `twingExtensions.js`) and `"loadExtensionsFile": "twingExtensions.js"`
 - this JS file must export a Map for `TwingEnvironment.addExtensions(extensions: Map<string, TwingExtensionInterface>)`
 - this Map will be added to the TwingEnvironment on startup

----

### Example

```js
// twingExtensions.js
const { TwingExtension, TwingFunction } = require('twing');

const extensionsMap = new Map();

class TestTwingExtension extends TwingExtension {
  getFunctions() {
    return [
      new TwingFunction('foobar', function (foo) {
        return Promise.resolve(`function foobar called with param "${foo}"`);
      }),
    ];
  }
}
extensionsMap.set('TestTwingExtension', new TestTwingExtension());

module.exports = extensionsMap;
```

See https://nightlycommit.github.io/twing/advanced.html#creating-an-extension for more details on how to create extensions


JosefBredereck:
That is an exciting extension of the pattern lab workspace. I want to request additional documentation for the parameter and what the extension file must contain.

@mfranzke, what do you think? Where is the best place to put this documentation?

1. https://github.com/pattern-lab/patternlab-node/blob/dev/packages/docs/src/docs/advanced-config-options.md
2. https://github.com/pattern-lab/patternlab-node/blob/dev/packages/engine-twig/README.md

mfranzke:
> That is an exciting extension of the pattern lab workspace. I want to request additional documentation for the parameter and what the extension file must contain.
> 
> @mfranzke, what do you think? Where is the best place to put this documentation?
> 
> 1. https://github.com/pattern-lab/patternlab-node/blob/dev/packages/docs/src/docs/advanced-config-options.md
> 2. https://github.com/pattern-lab/patternlab-node/blob/dev/packages/engine-twig/README.md

@JosefBredereck Valid question. I would say that it's important to mention `engines -> twig -> loadExtensionsFile` within the [advanced config options documentation](https://github.com/pattern-lab/patternlab-node/blob/dev/packages/docs/src/docs/advanced-config-options.md), to keep a general overview in there regarding the configurations. But put the rest into <https://github.com/pattern-lab/patternlab-node/blob/dev/packages/engine-twig/README.md> to keep it all in one place there regarding "installation" itself most likely with a hyperlink to [advanced config options documentation](https://github.com/pattern-lab/patternlab-node/blob/dev/packages/docs/src/docs/advanced-config-options.md) as well.

JosefBredereck (approved):

JosefBredereck:
@hadl do you want to add the documentation to this PR or in another one?

hadl:
docs added as @mfranzke suggested

PR #1459, opened 2022-12-27
refactor: enable fileendings separated by more than one dot

### LINKED ISSUE #1030: Fully support .html.twig suffix
I am using Pattern Lab Node `v3.0.1` on `macOS`, with Node `v10.6.0`, using the `Twig` Edition.

##### Expected Behavior

Use `.html.twig` suffix for patterns without any problems.

This style of suffix is advocated in [Symfony documentation](https://symfony.com/doc/current/templating.html#template-suffix) and is also used by Drupal. In some cases, e.g. when [registering layout templates](https://www.drupal.org/docs/8/api/layout-api/how-to-register-layouts), Drupal expects the layout template to have a `.html.twig` suffix.

##### Actual Behavior

Twig template source/rendered HTML not shown in UI, possible other problems.

##### Steps to Reproduce

Use `.html.twig` suffix for patterns.

mfranzke:
Closes https://github.com/pattern-lab/patternlab-node/issues/1030

### Summary of changes:
- enhancing fileextensions list in `packages/engine-twig/lib/engine_twig.js` and `packages/engine-twig-php/lib/engine_twig_php.js` to even also include `.html.twig` besides `.twig`
- rework in `packages/core/src/lib/object_factory.js` to check for a configured multi-dotted file extension and remove this form one from variable `this.fileName` whereas replace it in `this.fileExtension`


packages/engine-handlebars/lib/engine_handlebars.js
```diff
@@ -129,8 +129,8 @@ const engine_handlebars = {
    * assume it's already present
    */
   spawnMeta: function (config) {
-    this.spawnFile(config, '_head.hbs');
-    this.spawnFile(config, '_foot.hbs');
+    this.spawnFile(config, '_head.' + config.patternExtension);
```
JosefBredereck, line 127: This could be dangerous.
When pattern lab discovers multiple pattern engines, it will try to spawn the file found in the engine's folder located under `_meta/head...`
In that case, the file will not be found, and an error will be thrown here https://github.com/pattern-lab/patternlab-node/blob/4cf86b610721b55dec8390d08771528dbc20952d/packages/engine-handlebars/lib/engine_handlebars.js#L116
You can either catch the error and ignore it, or you revert it to its former state.

Another issue could be, that by chance the file will be taken, but the output for the twig file will be overwritten by the content of the `hbs` file. In that case, it could lead to an error when serving pattern lab because it has a different rendering engine.

JosefBredereck: I would rather look into the `engineFileExtensions` and when the `config.patternExtension` is included there, output the required file. (In that case, we need to read the file we deliver with the engine from the _meta folder and spawn it as the configured extension in `config.patternExtension`.

JosefBredereck (changes requested):

PR #1483, opened 2023-02-12
feat: migrate to nx workspace

JosefBredereck:
# Summary of changes:

1. migrate to Nx workspace
2. use Nx build scripts
3. use Nx testing behavior
4. introduce TypeScript
5. introduce new `@pattern-lab/types` package for common types over all packages
6. rework project structure
7. remove outdated packages (`uikit-polyfills`, `engine-react`, `development-edition-engine-react`)
8. added `development-edition-engine-twig-php`

# Open Points

- [ ] packages currently cannot be published as one workspace release until https://github.com/jscutlery/semver/issues/61 is resolved
- [ ] package dependencies in generated package.json files are excessive, which resolve in multiple open issues https://github.com/nrwl/nx/issues/10227#issuecomment-1427064452
- [ ] The info on how to cut a release needs to be updated
- [ ] Types package only includes rough base types and needs to be extended
- [ ] pre-commit hooks have to be reactivated
- [x] find a viable solution to use the precompiled sources and packages in development editions
- [ ] view all pages are not rendered and have to be fixed
- [ ] core and CLI tests have to be run-able


mfranzke:
> remove outdated packages (uikit-polyfills, engine-react, development-edition-engine-react)

Is removing `engine-react` and related packages very important to you? I could think about that actually updating and finalizing these react related packages instead would heavily benefit our popularity and make pattern lab even further attractive by new users.

JosefBredereck:
> > remove outdated packages (uikit-polyfills, engine-react, development-edition-engine-react)
> 
> Is removing `engine-react` and related packages very important to you? I could think about that actually updating and finalizing these react related packages instead would heavily benefit our popularity and make pattern lab even further attractive by new users.

I know what you want to imply by that, and it surely would improve our overall popularity. But I don't see any of us working on that feature and bring it to the point storybook did. Where you can simply add storybook to your workspace with one command and just include your component.
Therefore, I think we should focus on the static HTML generation engines and not the frameworks. This will give us more time for maintenance and improvements on that front.

PR #1489, opened 2023-02-25
docs: corrected that entry in the specs

### LINKED ISSUE #1480: [Documentation] Engine configuration fileExtension vs extension
I use Pattern Lab Node `v6.0.1` on `Windows | Mac | Linux`, with Node `v16.15.1`, using a `Vanilla` Edition with a little bit of gulp flavoured on top.

I recently upgrade to version 6.0.1 everything works as expected but I got an error message that the engine is not configured correctly. The engine configuration looked as follows:

```json
{
  "engines": {
    "handlebars": {
      "package": "@pattern-lab/engine-handlebars",
      "extensions": [
        "handlebars",
        "hbs"
      ],
      "extend": "helpers/hbs/*.js"
    }
  }
}
```

This is according to the documentation: [Engine](https://patternlab.io/docs/editing-the-configuration-options/#heading-engines)
However, I created a new project based on version 6.0.0.

The configuration in 6.0.0 looks as followed:

```json
{
  "engines": {
    "handlebars": {
      "package": "@pattern-lab/engine-handlebars",
      "fileExtensions": [
        "handlebars",
        "hbs"
      ],
      "extend": "helpers/hbs/*.js"
    }
  }
}
```

After I modified my configuration to "fileExtension" instead of "extension", the error message was not showing up anymore. So the question is now. Is the documentation outdated, which it seems to be, or an issue in pattern lab that got introduced in v6.0.1

mfranzke:
Closes #1480

### Summary of changes:
Corrected `extensions` to be `fileExtensions` in the documentation.

### DIFF
--- a/packages/docs/src/docs/advanced-config-options.md
+++ b/packages/docs/src/docs/advanced-config-options.md
@@ -198,7 +198,7 @@ Other engine specific configuration options can be added and will be passed to t
   "engines": {
     "handlebars": {
       "package": "@pattern-lab/engine-handlebars",
-      "extensions": [
+      "fileExtensions": [
         "handlebars",
         "hbs"
       ],


JosefBredereck (approved):

JosefBredereck:
Even if it currently doesn't make a difference, but it is a good future adjustment.

